PR #195, opened 2015-06-25
dev document...

zmathe:


atsareg:
Not sure I understand this commit. Was it meant to go here and not to DIRACDocs ?

zmathe:
well this is the doc which helps the developers. I do not think it can go to DIRACDocs. These files are generated by JSDUCK. Maybe it is better to write a scrip which generate the docs. I would like to have it because it is a very nice and helps others. You can have a look:
http://lhcb-cert-dirac.cern.ch/DIRAC/static/docs/index.html
Maybe it is better to write a script which create this document on-demand...

zmathe:
http://lhcb-cert-dirac.cern.ch/DIRAC/static/docs/index.html#!/api/Ext.dirac.utils.DiracJsonStore

atsareg:
So, this is the doc produced from the code itself ? It means that each time you change the code this should be regenerated ? If this is the case, you should certainly produce a script to generate it in place.
For example, at the moment of the new version installation

zmathe:
Yes, it is produced using the code. It is like doxygen. Each time when the code have changed this document has to be regenerated. 
I will write a script which generates all the files.

PR #200, opened 2015-07-08
improvements

zmathe:


WebApp/static/core/js/views/tabs/Main.js
```diff
@@ -1892,6 +1894,32 @@ Ext.define('Ext.dirac.views.tabs.Main', {
         }
 
       },
+      __defferOpenApplication : function(activeTab, tab) {
```
atsareg: English orthography is not yet abolished

PR #204, opened 2015-07-27
Modified SysAdmin page and added some doc

Sbalbp:


Sbalbp:
Modified SysAdmin page to enable upgrading hosts to newer versions of DIRAC from it

zmathe:
I found a small bug, which I have fixed:
https://github.com/DIRACGrid/WebAppDIRAC/pull/207

PR #215, opened 2015-10-08
PATCH: real root path to enabling module web extensions

vmendez:
I saw in dirac.egi.eu VMDIRAC module in new web was not loaded:
2015-10-08 08:26:51 UTC WebApp/Routing   INFO: Found handler WebApp.handler.VMDiracHandler
so the VM extension is correctly obtained

WARNING:tornado.general:403 GET /DIRAC/s:EGI-Production/g:breakseq_cloud/static/VMDIRAC/VMDirac/classes/VMDirac.js (85.58.20.119): /opt/dirac/pro/VMDIRAC/WebApp/static/VMDIRAC/VMDirac/classes/VMDirac.js is not in root static directory
2015-10-08 08:28:45 UTC WebApp/Web   WARN: 403 GET /DIRAC/s:EGI-Production/g:breakseq_cloud/static/VMDIRAC/VMDirac/classes/VMDirac.js (85.58.20.119) 1.20ms

403 is a forbiden error for this file.

I just do and test this hotpatch to take real paths, and it is working

### DIFF
--- a/Core/HandlerMgr.py
+++ b/Core/HandlerMgr.py
@@ -36,9 +36,11 @@ class HandlerMgr( object ):
         continue
       try:
         modFile, modPath, desc = imp.find_module( extName )
+	#to match in the real root path to enabling module web extensions (static, templates...)
+	realModPath = os.path.realpath( modPath )
       except ImportError:
         continue
-      staticPath = os.path.join( modPath, "WebApp", dirName )
+      staticPath = os.path.join( realModPath, "WebApp", dirName )
       if os.path.isdir( staticPath ):
         pathList.append( staticPath )
     #Add WebAppDirac to the end


zmathe:
The real problem is different... You must get the correct path modFile, modPath, desc = imp.find_module( extName )!
LHCbWebAppDIRAC is an extension of WebAppDIRAC and it works. Where I can check your configuration and also the code?

zmathe:
okay! I found the code. You have VMDIRAC/WebApp/static/VMDIRAC/VMDirac
You must have VMDIRAC/WebApp/static/VMDIRAC/
You do not need the second VMDirac

vmendez:
The real problem is tornado is taking root static path in:
 /opt/dirac/versions/v6r14p5_1444148305
but not in /opt/dirac/pro even when this one is a link of the first

This patch resolves the issue and it would be also valid for your LHCbWebAppDIRAC
would you pull to integration, please ?

zmathe:
I do not think: /opt/dirac/pro/VMDIRAC/WebApp/static/VMDIRAC/VMDirac/classes/VMDirac.js is not in root static directory
The directory structure is wrong! We have decided:
VMDIRAC/WebApp/static/VMDIRAC
What you have 
VMDIRAC/WebApp/static/VMDIRAC/VMDirac! This is wrong.
I have a feeling that will breaks the compilation. 
Andrei, merge the PRs.

atsareg:
So, what is the conclusion. Is it a correct change in the code ? I am bit confused by this discussion

zmathe:
It is a correct change. But it may break something else...
As I said before the extension must be: VMDIRAC/WebApp/static/VMDIRAC
I strongly suggest to change VMDIRAC/WebApp/static/VMDIRAC/VMDirac to VMDIRAC/WebApp/static/VMDIRAC.

vmendez:
Zoltan the VMDIRAC/WebApp/static/VMDIRAC/VMDirac question is useless in this PR.
You can put what ever you want, but if it is out of tornado static root path, then it will never work.
Whether it is working in your LHCbWebAppDIRAC is just because in your static and template extensions are within tornado static root path (the real path not the link), then this patch will also works because  realModPath = modPath and nothing change for you
modFile, modPath, desc = imp.find_module( extName )
realModPath = os.path.realpath( modPath )

zmathe:
so, You can also put your staff under tornado static root path... Could you please try to execute  dirac-web-compile?
Your modification does not change nothing in my cases.

vmendez:
Considering it does not change anything in your cases, then Andrei could you pull to integration, please ?

atsareg:
I merge the PR. @vmendez However please update the code structure of VMDIRAC/WebApp to the standard one and validate the change

PR #221, opened 2016-01-29
Extensions' information

Sbalbp:
Modified the SysAdmin page to display information about the versions of the extensions installed in hosts ( DIRACGrid/DIRAC#2785 ).


WebApp/handler/SystemAdministrationHandler.py
```diff
@@ -73,6 +73,12 @@ def web_getSysInfo( self ):
       return
     
     callback = sorted(callback, key=lambda i: i['Host'])
+
+    # Add the information about the extensions' versions, if available, to display along with the DIRAC version
+    for i in range( 0, len( callback ) ):
```
fstagni: range(0, x) is same as range(x)

fstagni:
Review OK

PR #224, opened 2016-02-04
job summary page implemented

zmathe:


andresailer:
Thanks for implementing this

zmathe:
I was thinking to have a possibility to open the SiteSummary application from the JobSummary:
Right click on a site and you will have possibility to open the JobMonitor and also SiteSummary.
Do you think this is important?

Issue #230, opened 2016-03-16
Sytem Administration App Issues


andresailer:
Some issues observed in the SystemAdministrator app:
- [x] 1. If the header entries are clicked to sort the list of machines a second time, the columns are no longer aligned with the fields...
   ![sysadmin1](https://cloud.githubusercontent.com/assets/2115797/13813201/8a702adc-eb80-11e5-811f-e9cf37a3c803.png)
   
   Also it seems that the whole webapp crashes. No more tabs can be opened once this happened.
- [x] 2. Sorting the services/agents in the lower part of this app is not working at all. Nothing happens when the header fields are clicked.
- [x] 3. The Restart/Start/Stop buttons in the lower half don't work.
   Selecting agents and then clicking restart only gives the error message:

```
Failed to restart Components:
voilcdirac032.cern.ch: Can't connect to dips:// voilcdirac032.cern.ch:9162/Framework/SystemAdministrator: {'Message
': "Could not resolve voilcdirac032.cern.ch: Can't get info for host voilcdirac032.cern.ch: [Errno -2] Name or service not known", 'OK': False}

voilcdirac032.cern.ch: Can't connect to dips:// voilcdirac032.cern.ch:9162/Framework/SystemAdministrator: {'Message': "Could not resolve voilcdirac0
32.cern.ch: Can't get info for host voilcdirac032.cern.ch: [Errno -2] Name or service not known", 'OK': False}
```

Another three issues:
- [x] In the "Show errors" window (the window popping up for host -> right click -> show errors) sorting is also not possible. default sort of "errors per hour" would be great
- [x] In the same error overview window one can no longer go straight to the log of the agent/service (right click -> Show log was possible in the old webinterface)
- [x] The window showing the log file does not have a "Refresh" button any longer.

andresailer:
Another two issues:
- [ ]  In the "Show errors" window (the window popping up for host -> right click -> show errors) sorting is also not possible. default sort of "errors per hour" would be great
- [ ]  In the same error overview window one can no longer go straight to the log of the agent/service (right click -> Show log was possible in the old webinterface)

zmathe:
1->fixed in #282

zmathe:
2->fixed

PR #240, opened 2016-06-07
Fix Restart button on SysAdministration toolbar

sfayer:
This fixes a bug we found where the Restart button on the System Administrator (component) toolbar doesn't work (and possibly the other actions too), but right clicking on an individual item does. This is caused by an additional space on the end of the component name when the request is sent to the server.

Regards,
Simon


zmathe:
Review OK

andresailer:
This fixes issue number 3 of #230 (first comment) , right? Thanks.

sfayer:
Yes, that was the same error that we were seeing (bizarre unknown host errors due to the extra space messing up the resolver)...
I think the other problems from that issue can probably be fixed by setting remoteSort = false on the two tables, but I don't know if it would cause any other problems.

PR #245, opened 2016-07-19
using futures ThreadPoolExecutor instead of DIRAC ThreadPool

zmathe:
This requires the futures library.

### DIFF
--- a/Lib/WebHandler.py
+++ b/Lib/WebHandler.py
@@ -1,6 +1,5 @@
 
 from DIRAC import gLogger
-from DIRAC.Core.Utilities.ThreadPool import getGlobalThreadPool
 from DIRAC.Core.Security.X509Chain import X509Chain
 from DIRAC.Core.DISET.ThreadConfig import ThreadConfig
 from DIRAC.ConfigurationSystem.Client.Helpers import Registry
@@ -19,6 +18,7 @@ import tornado.ioloop
 import tornado.gen
 import tornado.stack_context
 import tornado.websocket
+import concurrent.futures
 
 class WErr( tornado.web.HTTPError ):
   def __init__( self, code, msg = "", **kwargs ):
@@ -52,7 +52,7 @@ def asyncGen( method ):
 
 class WebHandler( tornado.web.RequestHandler ):
 
-  __threadPool = getGlobalThreadPool()
+  __threadPool = concurrent.futures.ThreadPoolExecutor( 1000 )
   __disetConfig = ThreadConfig()
   __log = False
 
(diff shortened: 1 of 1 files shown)


zmathe:
This is related to #246

fstagni:
It seems even too easy to be true. Now, for my info: this change just uses the `cuncurrent` library for the threadpool. What about now creating more than one process (so that the blocking I/O is exploited)?

zmathe:
Well, it toke me more than one week, to found out what is the problem. Okay. I wanted also to not use Adri sparta but I was not able to pass the user certificate. So when the user triggered a call from the web portal (for example browsing the bookkeeping) then in the web server the user authenticated with his user DN, but I was not able to pass the user DN to the RPC call (it means the machine cert is used). So I had to use the famous sparta.
It works well with blocking I/O and this is only one which works using more than one processes.

fstagni:
Just basic question: where do you get it? I see it's present on LHCb certification machine, but not in other installations, like:

```
[dirac@lbvobox100 pro]$ python
Python 2.7.8 (default, Apr 28 2015, 23:38:39) 
[GCC 4.4.7 20120313 (Red Hat 4.4.7-11)] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> import concurrent
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
ImportError: No module named concurrent
```

On my ubuntu installation:
```
(tryEmpty)  ✘ toffo@pclhcb181  ~/pyDevs/LHCbDIRAC   fixesv8r3_3  python
Python 2.7.6 (default, Jun 22 2015, 17:58:13) 
[GCC 4.8.2] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> import concurrent.futures
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
ImportError: No module named concurrent.futures
```

On lxplus:
[LHCbDirac v8r3-pre23] ~ $ python
Python 2.7.10 (default, Mar 10 2016, 17:36:04) 
[GCC 4.9.3] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> import concurrent.futures
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
ImportError: No module named concurrent.futures
>>>

atsareg:
You have to do 
pip install futures
before using it with python 2.7. I guess this is what Zoltan did

zmathe:
I have installed to certification that is why you have in certification

fstagni:
Ok I see, thanks. Anyway I don't see `futures` in the Externals, and 
this is a pre-requisite.

zmathe:
This PR requires #3047. It can not be deployed before...

PR #251, opened 2016-08-03
Prevent XSS

vipersec:
After a little bit of investigation I found that the dirac web application is vulnerable to [reflected cross site scripting attacks (XSS)](https://en.wikipedia.org/wiki/Cross-site_scripting).

XSS is a very common vulnerability which is often used to hijack a user's session.

The following parameters were affected: `url_state`, `view`, `open_app`
Here is a screenshot that confirms the existence of the vulnerability:
xss2

I added a filter that removes the `<` `>` tags using regular expressions.
Feel free to discuss the changes.

### DIFF
--- a/WebApp/handler/RootHandler.py
+++ b/WebApp/handler/RootHandler.py
@@ -5,6 +5,11 @@ from WebAppDIRAC.Lib import Conf
 from WebAppDIRAC.Lib.WebHandler import WebHandler, WErr
 import re
 
+def xss_filter(text):
+  cleanr =re.compile('<.*?>')
+  cleantext = re.sub(cleanr,'', text)
+  return cleantext
+
 class RootHandler(WebHandler):
 
   AUTH_PROPS = "all"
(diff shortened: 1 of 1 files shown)


zmathe:
The attacker need a valid proxy. The protection from the XSS attack is the proxy. The example what you showed me is modified the code in your browser, but it does not affect to me or other users. These parameters are not stored in the server side.

vipersec:
@zmathe 
I assume that the valid proxy is the certificate in our browser which is provided by Cern.
This only stops an unauthorized user from accessing the site, right? It's true that it stops an attacker but not fully.

Once someone has any certificate of a normal he can easily steal the session of an administrator.
The fact that any authorized user can get the session of any other user is an issue in my opinion.

Yes, in my example I'm injecting javascript so I can access the cookies of the site in my browser, this is only a proof of concept. 

In a real scenario the attacker sends this URL to a potential victim (but instead of using `alert` he stores the victim's cookie in a server that he controls)
After getting the victim's cookie he can place it in his browser to impersonate the victim.

PR #258, opened 2016-10-14
[v6r16] [WebApp =v2r0] make the Acounting and Monitoring more general...

zmathe:


fstagni:
This PR contains tons of changes. I guess this is for new release only. 2 questions:
- do you want the tag that includes this to be called v1r7 or v2r0?
- should this come with DIRAC v6r16? I guess so (the alternative is to wait for DIRAC v6r17)

zmathe:
This will be v1r7. This will come with v6r16, it can not wait as this is the Monitoring application. Without this we can not see the plots...

zmathe:
This will be a new major release v2r0 as google.visualization api is no longer used...

fstagni:
I am wondering if WebApp extensions will need to reflect the changes done here, or if it would be transparent. I guess "it depends", but can you be precise about that?

zmathe:
What do you mean? which changes?

fstagni:
Just to say an example, if I search for "google" in LHCbWebAppDIRAC I find an entry in RAWIntegrityMonitor.js. Should it be moved to using ExtJS charts?

zmathe:
Yes, because the https://gitlab.cern.ch/lhcb-dirac/LHCbWebDIRAC/merge_requests/10/ is not merged. I have replaced here also. But anyway I was not able to remove the dependency as we have a widget which does not have any equivalent in Extjs 4. But in ExtJs 6 I will be able to remove the google dependency.

fstagni:
Ok for LHCb, my question is mostly for other VOs: do we need to ask to them to intervene on their extensions? Or is it enough to say that we advise them to do this intervention?

zmathe:
I would say I just write a message that say WebApp will not support Google visualization api from v3r0 (this will be based on ExtJs 6). And we can advertise in the next Bild-Dev meeting as well.
ExtJs 6 has an integrated google visulization api as well. But if it is used external google links, we can not use it.

fstagni:
Yes please add a message in https://github.com/DIRACGrid/DIRAC/wiki/DIRAC-v6r16

PR #264, opened 2016-11-29
[v2r0p5]take into account the report type selector

zmathe:


zmathe:
@atsareg , @fstagni please include to the new release

PR #265, opened 2016-12-06
[v2r1] allow hyphen in the file name and correct the accounting handler

zmathe:


zmathe:
This is for v6r16 and v6r17!

PR #280, opened 2017-10-16
Add welcome window

TaykYoku:


WebApp/static/core/js/views/desktop/Main.js
```diff
@@ -2473,3 +2473,14 @@ Ext.define('Ext.dirac.views.desktop.Main', {
     /*-----------------END - IMPLEMENTATION OF THE INTERFACE BETWEEN STATE MANAGEMENT ADN DESKTOP----------------*/
 
   });
+  /*
+   * Welcome window
+   */
+  if (GLOBAL.APP.configData.user.username) {
```
zmathe: This is not following the ExtJS developer convention. I know it is JS and this code can be executed everywhere, but you must add this code inside a class. What is a class in ExtJs:

Ext.define('Ext.dirac.views.desktop.Main', {
  extend : 'Ext.panel.Panel',
.....
});
In addition I do not like it at all. Every time when I open or refresh my browser I get this message. I have to click to close. Most of the users know what is DIRAC.  You also added a hyper link, so if someone click on it the actual web page will be overwritten.
Conclusion:
-In this format can not be approved. You should add this few lines inside a class.
-I do not like this features, I would say we must discuss this in the BILD meeting.

atsareg:
The idea is that this Welcome panel is shown when a Visitor accesses the portal. Authenticated user should not see it. Also just reloading the portal page should not trigger the Welcome message. I guess this should be possible to do. @zmathe: please suggest the appropriate solution. This Welcome splash screen is a strong requirement for multi-VO installations. It should be possible to disable it by configuration. The contents of the panel should be easily customizable.

WebApp/template/root.tpl
```diff
@@ -141,9 +141,28 @@
...
           </tr>
         </table>
+      </div>
+  </div>
 
+  <div id="app-dirac-welcome">
+      <div class="app-dirac-welcome-text">
+        <h1>Weclome to DIRAC!</h1>
+        <b>D</b>istributed <b>I</b>nfrastructure with <b>R</b>emote <b>A</b>gent <b>C</b>ontrol INTERWARE is a software framework for distributed computing providing a complete solution to one (or more) user community requiring access to distributed resources. DIRAC builds a layer between the users and the resources offering a common interface to a number of heterogeneous providers, integrating them in a seamless manner, providing interoperability, at the same time as an optimized, transparent and reliable usage of the resources. 
+        <a href="http://diracgrid.org">Discover more...</a>
```
zmathe: Better to open the new link in a new tab. Could you please use: <a target="_blank" href="http://diracgrid.org">Discover more..</a> ?

WebApp/static/core/js/views/tabs/Main.js
```diff
@@ -2371,3 +2371,14 @@ Ext.define('Ext.dirac.views.tabs.Main', {
...
+ // } else {
+    /*
+     * If the user is not registered
+     */
+  //  Ext.Msg.alert('Welcome to DIRAC!', "Distributed Infrastructure with Remote Agent Control INTERWARE is a software framework for distributed computing providing a complete solution to one (or more) user community requiring access to distributed resources. DIRAC builds a layer between the users and the resources offering a common interface to a number of heterogeneous providers, integrating them in a seamless manner, providing interoperability, at the same time as an optimized, transparent and reliable usage of the resources. <br> <i><a href='http://diracgrid.org'>Discover more...</a></i> <br><br> <b>You are not registry user.</b> <br> If you want to use a user proxy, we assume that you already have a user certificate, so in this case create a directory .globus in your home directory and copy the certificate files (public and pruvate keys in pem format) to this directory: <br> $ mkdir ~/.globus <br> $ cp <<certificate files>> ~/.globus/ <br> At this point you need a proxy, you should do: <br> $ dirac-proxy-init");
+  //}
```
zmathe: Could you please remove the commented lines? Please add this line Ext.create('Ext.dirac.views.desktop.WelcomeWindow'); also to the Ext.dirac.views.tabs.Main

zmathe:
please apply  what I requested.  The commented lines are still there and the WelcomeWindow is not in Ext.dirac.views.tabs.Main

PR #281, opened 2017-10-17
New distribution

zmathe:
This is part of #3498


zmathe:
@atsareg Could you please include to the new pre release?

PR #283, opened 2017-12-20
Removed RequestMonitor

fstagni:
This PR simply comments out the RequestMonitor web page. The reason is that it is a "dangerous" Web App: though that one can potentially launch extremely heavy requests to RequestDB. 

I am totally aware that there are much better ways: the problem is not the App itself but what you can do with the DB (and with the RequestManager service), so this is just a quick'n dirty solution. I have created an issue in https://github.com/DIRACGrid/DIRAC/issues/3581


zmathe:
We do not have to remove, because other VOs may use it. We will create our own configuration file and do not add there....

fstagni:
So, shall we simply close this PR?

zmathe:
Yes, I close it and going to prepare our own configuration....

PR #285, opened 2018-01-23
[v3r1p2] allow to configure backround and logo images

zmathe:


WebApp/handler/RootHandler.py
```diff
@@ -90,11 +90,15 @@ def web_index(self):
       open_app = xss_filter( self.request.arguments["open_app"][0].strip() )
 
     icon = data[ 'baseURL' ] + Conf.getIcon()
-
+    backround = data[ 'baseURL' ] + Conf.getBackround()
```
fstagni: `background` not `backround` :)

Lib/Conf.py
```diff
@@ -136,3 +136,9 @@ def SSLProrocol():
 
 def getStaticDirs():
   return getCSValue( "StaticDirs", [] )
+
+def getLogo():
+  return getCSValue("Logo","/static/core/img/icons/system/_logo_waiting.gif")
+
+def getBackround():
```
fstagni: `getBackgroud`

WebApp/template/root.tpl
```diff
@@ -82,6 +82,7 @@
           GLOBAL.VALID_VIEWS = ["desktop","tabs"];
           GLOBAL.MAIN_VIEW_SAVE_STRUCTURE_VERSION = 1;
           GLOBAL.OPEN_APP = "{{open_app}}";
+          GLOBAL.BACKGROUND = "{{backroundImage}}";
```
fstagni: and here.

WebApp/template/root.tpl
```diff
@@ -113,6 +114,7 @@
...
```
fstagni: and here

Lib/Conf.py
zmathe: Done

WebApp/static/core/js/views/tabs/RightContainer.js
```diff
@@ -416,7 +416,7 @@ Ext.define('Ext.dirac.views.tabs.RightContainer', {
                   isLoaded : isLoaded,
                   bodyStyle : {
                     background : '#AAAAAA',
-                    backgroundImage : 'url(/DIRAC/static/core/img/wallpapers/dirac_background_6.png)',
+                    backgroundImage : 'url('+GLOBAL.BACKGROUND+')',
```
fstagni: This is mostly my ignorance with js, so can you explain why at this line you used `+GLOBAL.BACKGROUND+` and at line 83 only `GLOBAL.BACKGROUND` ?

zmathe: In the line 83 a setter method is used. The wallpaper is a widget...  The setter method is doing all required rendering, etc. 
+GLOBAL.BACKGROUND+ is used inside a custom CSS property (bodyStyle) as you can see in line 417.  I set the background using CSS element. You can see how you can set a background using CSS: https://www.w3schools.com/cssref/pr_background-image.asp

PR #288, opened 2018-03-20
[v3r1p5] allow to set authorisation roles for a certain request

zmathe:


zmathe:
@fstagni Could you please review this is part of #3643? @atsareg Could you please include to the new pre release?

PR #290, opened 2018-04-18
[v3r1p7] fix typo

zmathe:


zmathe:
@atsareg Could you please also include to v6r20 series?

PR #292, opened 2018-05-18
[v3r1] Fixed bug in getSelectionData

atsareg:
Minor bug fix


atsareg:
Actually already fixed but not yet released

PR #293, opened 2018-05-30
[v3r1p8] keep backward compatible for DIRAC v9r19

zmathe:


zmathe:
@atsareg Could you please include to v6r20p2?

PR #296, opened 2018-07-23
[v3r1] Nginx returns user DN not in the standard form

atsareg:
Nginx (sometimes ?) returns the user DN not in the standard form with fields separated by slashes.


Lib/WebHandler.py
```diff
@@ -163,6 +163,10 @@ def __processCredentials(self):
       headers = self.request.headers
       if headers['X-Scheme'] == "https" and headers['X-Ssl_client_verify'] == 'SUCCESS':
         DN = headers['X-Ssl_client_s_dn']
+        if not DN.startswith('/'):
```
zmathe: which kind of configuration do you have? Why we do not see?

zmathe:
I think so. Probably the DN is correct in LHCb use case.

PR #305, opened 2018-11-21
Make Welcome page configurable

TaykYoku:
Make Welcome page configurable

BEGIN

*Lib/Conf.py
FIX: Added dict with welcome page parameters

*WebApp/handler/RootHandler.py
FIX: Added dict with welcome page parameters

*WebApp/static/core/css/css.css
FIX: Deleted not needed options

*WebApp/static/core/js/utils/WelcomeWindow.js
FIX: Added creation of html elements

*WebApp/template/root.tpl
FIX: Deleted Welcome page html code and added dict to global variable

*WebApp/web.cfg
FIX: Added comments about Welcome options

END

### DIFF
--- a/Lib/Conf.py
+++ b/Lib/Conf.py
@@ -151,3 +151,13 @@ def getLogo():
 
 def getBackgroud():
   return getCSValue("BackgroundImage","/static/core/img/wallpapers/dirac_background_6.png")
+
+def getWelcome():
+  res = {}
+  res['show'] = getCSValue("WelcomePage/show", "False")
+  res['style'] = getCSValue("WelcomePage/style", "")
+  res['title'] = getCSValue("WelcomePage/title", "")
+  res['text'] = getCSValue("WelcomePage/text", "")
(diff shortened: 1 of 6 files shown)


Lib/Conf.py
```diff
@@ -150,4 +150,14 @@ def getLogo():
...
-  return getCSValue("BackgroundImage","/static/core/img/wallpapers/dirac_background_6.png")
\ No newline at end of file
+  return getCSValue("BackgroundImage","/static/core/img/wallpapers/dirac_background_6.png")
+
+def getWelcome():
+  res = {}
+  res['show'] = getCSValue("WelcomePage/show", "False")
+  res['style'] = getCSValue("WelcomePage/style", "")
+  res['title'] = getCSValue("WelcomePage/title", "")
+  res['text'] = getCSValue("WelcomePage/text", "")
```
atsareg: Putting large text into the CS is not very practical. I suggest that the returned string is checked if it is a name of existing file (e.g. /opt/dirac/etc/welcome.html) and if it is, read the text from the file.

PR #311, opened 2019-01-11
[v4r0] Fixed unauthorized error

TaykYoku:
Fixed "Unauthorized." 401 error in case cert is valid but not registered

BEGINRELEASE

*Lib/WebHandler.py
FIX: Filled credDict after checking DN

ENDRELEASE


zmathe:
You already made this modification, which broke other functionality. The authorisation of users is done later. Please check one of my merged PR.

PR #312, opened 2019-01-24
[v3r1p16]do not convert the URL to https when it is already https

zmathe:


fstagni (approved):

fstagni:
@atsareg don't forget this for the next release or pre-release

PR #316, opened 2019-05-21
fixes for missing CS module (for v6r22)

fstagni:
I also added the checks done by Travis and the PR template.

BEGINRELEASENOTES

CHANGE: Using PilotManager instead of WMSAdministrator
FIX: Using Clients instead of RPCClient
FIX: Removed AccountingPlotOldHandler

ENDRELEASENOTES


fstagni:
I also added Travis checks

zmathe:
We are not using the AccountingPlotOldHandler.py, you can delete it

PR #319, opened 2019-06-04
[v4r0] update FileSaver and use blob to save the file

zmathe:
BEGINRELEASENOTES

ConfigurationHandler
FIX: Use blob for saving the content of the CS.

ENDRELEASENOTES


fstagni (approved):

zmathe:
@atsareg Please merge this as well.  It is needed for v6r22

PR #320, opened 2019-06-13
bugfixes

zmathe:
BEGINRELEASENOTES

Core:
FIX: Compare the handler property with None

ENDRELEASENOTES


fstagni (dismissed):

.tern-project
```diff
@@ -0,0 +1,20 @@
+{
```
fstagni: Is this file here on purpose?

zmathe: this is the esLint configuration file which is used by eclipse. I put it in purpose but probably each developer should configure? I can not decide to put it here or not :)

fstagni: If it's something specific to Eclipse then it shouldn't be here.

fstagni (approved):

fstagni (approved):

PR #335, opened 2019-07-18
[v4r0] add EsLint to travis

zmathe:
BEGINRELEASENOTES

NEW: Allowing ESLint in travis

ENDRELEASENOTES


fstagni:
Travis is failing because you are trying `yum` but the Travis runners are ubuntu

andresailer:
Also one should add another entry in the test matrix for "ESLint" instead of running ESLint for all other tests.
Or maybe set up matrix for multiple languages as described here:
https://stackoverflow.com/questions/27644586/how-to-set-up-travis-ci-with-multiple-languages

zmathe:
Thanks Andre for the hint.
I wanted to use what they propose: https://github.com/eslint/eslint
I will try your proposal as well. 
There is also an example what I checked:
https://github.com/adametry/gulp-eslint/blob/master/.travis.yml

andresailer:
Using matrix inside matrix doesn't work, this now essentially disables the pylint and format tests.


You can just add an additional step to the matrix , so that the final yml looks something like this
(There are also no pytests, so codecov doesn't make sense, dropping after_success)
```yml
dist: trusty
language: python
python:
  - "2.7"

env:
  matrix:
    - CHECK=pylint
    - CHECK=format

# command to install dependencies
install: 
  - "pip install --upgrade setuptools"
  - "pip install --upgrade pip"
  - "pip install -r requirements.txt"
# command to run tests
script:
  - export PYTHONPATH=${PWD%/*}
  - ls $PYTHONPATH
  - if [[ "${CHECK}" == "pylint"  ]];
    then travis_wait 30 .travis.d/runPylint.sh;
    elif [[ "${CHECK}" == "format" ]] && [[ "${TRAVIS_PULL_REQUEST}" != "false" ]];
    then git remote add GH https://github.com/DIRACGrid/WebAppDIRAC.git; git fetch --no-tags GH ${TRAVIS_BRANCH}; git branch -vv; git diff -U0 GH/${TRAVIS_BRANCH} | pycodestyle --diff;
    fi

matrix:
  include:
    language: node_js
    node_js: 8.10.0
    before_install:
      - echo `PWD`
    install:
      - npm install eslint --save-dev
    script:
      - node_modules/.bin/eslint WebApp
```

zmathe:
@andresailer It is working, it is read because there are lot of error messages. I started to correct but I have not finished yet.

zmathe:
Screenshot from 2019-07-19 18-17-51

zmathe:
@andresailer I have added your configuration file and it seams it is better. I have to fix the JS errors.

andresailer:
Thanks!

fstagni (approved):

PR #337, opened 2019-07-31
Fix a bug in making extension name

zhangxiaomei:
BEGINRELEASENOTES

FIX: When the length of extension name is 4, there is a problem to add "DIRAC" as an end.   

ENDRELEASENOTES


fstagni:
Thanks. Can you rebase it to v4r0?

zhangxiaomei:
Done.

fstagni (approved):

PR #341, opened 2019-08-15
[v4r0] various improvements

zmathe:
BEGINRELEASENOTES

Core:
FIX: fix the selector and improve the GridPanel

ENDRELEASENOTES


WebApp/static/core/js/utils/DiracBoxSelect.js
```diff
@@ -1,13 +1,21 @@
 Ext.define('Ext.dirac.utils.DiracBoxSelect', {
       extend : "Ext.form.field.Tag",
-      filterPickList : true,
+      // filterPickList : true,
+      scrollable : true,
+      resisable : true,
```
fstagni: Shouldn't it be `resizable`?

PR #342, opened 2019-08-18
[v4r1] Merge with v4r0

TaykYoku:
Merge with v4r0

BEGINRELEASENOTES

*Configuration
NEW: Add new case in logic to force refresh gConfig

ENDRELEASENOTES


fstagni:
Mmm... not sure why this PR is here... I think simply @atsareg can merge v4r0 in v4r1 like we always do?

PR #347, opened 2019-09-02
[v4r0] allow to select text from window

zmathe:
BEGINRELEASENOTES

FIX: allow to select text from window

ENDRELEASENOTES


fstagni:
Have all the commits been merged?

PR #355, opened 2019-10-22
[v4r0] SiteSummaryHandler inherits from ResourceSummaryHandler 

fstagni:
(implement missing functionalities in SiteSummaryHandler)

BEGINRELEASENOTES

CHANGE: SiteSummaryHandler inherits from ResourceSummaryHandler (implement missing functionalities in SiteSummaryHandler)

ENDRELEASENOTES

### DIFF
--- a/WebApp/handler/SiteSummaryHandler.py
+++ b/WebApp/handler/SiteSummaryHandler.py
@@ -7,45 +7,15 @@ from DIRAC.Core.Utilities.Plotting.FileCoding import codeRequestInFileId
 from DIRAC.ResourceStatusSystem.Utilities.CSHelpers import getSiteComputingElements
 from DIRAC.ResourceStatusSystem.Client.PublisherClient import PublisherClient
 
-from WebAppDIRAC.Lib.WebHandler import WebHandler, asyncGen
+from WebAppDIRAC.Lib.WebHandler import asyncGen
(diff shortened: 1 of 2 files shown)


WebApp/handler/SiteSummaryHandler.py
```diff
@@ -77,73 +47,6 @@ def web_getSiteSummaryData(self):
 
     self.finish(result)
 
-  @asyncGen
-  def web_action(self):
```
zmathe, line 81: Is this method is not used? Have you removed from the javascript?

WebApp/handler/SiteSummaryHandler.py
```diff
@@ -77,73 +47,6 @@ def web_getSiteSummaryData(self):
...
-      except AttributeError:
-        result = {'success': 'false', 'error': 'bad action %s' % actionName}
-
-    else:
-
-      result = {'success': 'false', 'error': 'Missing action'}
-
-    self.finish(result)
-
-  def _getHistory(self, requestParams):
```
zmathe, line 104: Is this method no longer used?

zmathe (approved):

WebApp/handler/SiteSummaryHandler.py
fstagni: These 2 methods are used, simply they inherit from ResourceSummaryHandler (the 2 pages are very similar indeed!)

PR #363, opened 2020-01-21
[v4r0] take into account the property for the main menu

zmathe:
BEGINRELEASENOTES

FIX: Allow to disable main menu if the user does not have the required DIRAC role.

ENDRELEASENOTES


zmathe:
@atsareg Please add this PR to the new release.

PR #364, opened 2020-01-24
[v4r1] Add possibility to load state to JobLaunchpad, small fixes

TaykYoku:
BEGINRELEASENOTES

NEW: For JobLaunchpad add loadState method and some logic from eiscat project to add possibility to load options from another application window
FIX:  fix raw to data for node, optimize, resolve some warnings

ENDRELEASENOTES

### DIFF
--- a/WebApp/static/DIRAC/ActivityMonitor/classes/ActivityMonitor.js
+++ b/WebApp/static/DIRAC/ActivityMonitor/classes/ActivityMonitor.js
@@ -1,1392 +1,1392 @@
 Ext.define('DIRAC.ActivityMonitor.classes.ActivityMonitor', {
-      extend : 'Ext.dirac.core.Module',
-      requires : ['Ext.util.*', 'Ext.panel.Panel', "Ext.form.field.Text", "Ext.button.Button", "Ext.menu.Menu", "Ext.form.field.ComboBox", "Ext.layout.*", "Ext.form.field.Date", "Ext.form.field.TextArea", "Ext.form.field.Checkbox", "Ext.form.FieldSet",
-          "Ext.dirac.utils.DiracMultiSelect", "Ext.toolbar.Toolbar", "Ext.data.Record", 'Ext.Array', 'Ext.data.TreeStore', "Ext.ux.form.MultiSelect", "DIRAC.ActivityMonitor.classes.ActivityTreeModel"],
+  extend : 'Ext.dirac.core.Module',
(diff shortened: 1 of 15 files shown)


fstagni:
@zmathe can you please review?

zmathe:
Well, he is not using the same code format (style). There is a recommended way to develop WebApp and there is a template to be used: https://github.com/DIRACGrid/WebAppDIRAC/blob/integration/extjs-template.xml  I do not know what is the reason to change the style... We should use the same code style which is defined in extjs-template.xml
However, the code point of view review is OK.

TaykYoku:
Thanks for fair comment on style. I'll try to configure my VScode and fix this

zmathe:
I am not strict with the format what we are using now. We can change it any time. But we have to decide the format. Probably, we can create a template for VScode as well? If all developer is using different code format, the code will be not nice.

zmathe:
@atsareg Why you closed this PR? In the code point of view is okay. We need to agree with the code style.

atsareg:
@zmathe: I had to reorganize rel-v4r1 and integration branches by hand bypassing the normal PR procedure. This was due to many conflicts some of which were resolved by replacing files with new versions. In particular, I had to recreate the rel-v4r1 branch which automatically triggered closing this PR. This is a one-time action and should not happen again in the future.
The fixes from this PR are integrated into the new rel-v4r1.

zmathe:
No, problem. But still the code style issue is valid. We have to found a solution which is good for all kind of IDEs.

atsareg:
Andrii tried to follow the style and not using the style of his IDE in the updated rel-v4r1. May be there are still discrepancies. So, a common style tool aka autopep8 for javascript would be needed. Do, you know of any tool like that ?

zmathe:
As far as I know there is no autopep8 for javascript. I have created a template which can be used by Eclipse, but it requires Eclipse. Probably, it is not idea to force developer to use Eclipse...
This is the documentation about setting up Eclipse: https://dirac.readthedocs.io/en/latest/DeveloperGuide/WebAppDIRAC/SetupEclipse/index.html
Probably, we need to expand and create an appropriate code style, which works not only with eclipse

fstagni:
Can you look at https://standardjs.com/ ?

TaykYoku:
as is being considered standardjs, it is also possible to consider https://prettier.io/, which is more popular

zmathe:
I had a look both proposal. prettier is only for code formatter standardjs is more. It is much much better. standardjs also has eslint. However, we already integrated eslint. So, taking into account prettier support more than one language I would say it is better. We have to consider that we will use JSON in the near future.

zmathe:
I just checked prettier and standardjs gives the same code style. So, we can start with prettier and if we will not like it, we still can replace...

PR #365, opened 2020-01-28
[WIP]move tarvis tests to github actions

zmathe:
BEGINRELEASENOTES

CHANGE: Use github actions

ENDRELEASENOTES


fstagni:
2 questions:
- why using flake8? I understand is (maybe?) better than pylint, but at the same time in the other DIRAC projects we prefer to use pylint instead of flake8, and I would use the same tools everywhere.
- You are using `virtualenv` and `pip`. I would suggest to use already `conda`, like it's done in https://github.com/DIRACGrid/DIRAC/blob/integration/.github/workflows/basic.yml#L40 as it will make easier to add more python versions at some point.

zmathe:
- I do not have any reason to use flake8. They complain it is better and I wanted to use. But I have no problem to use pylint.
- I am going to use conda

fstagni:
Looks OK. Would it be possible to backport (to rel-v4r0)?

zmathe:
Yes, I will squash all commits and cherry-pick.  I am still working on it. It turned out there is an esLint plugin for extjs and try to use it.

zmathe:
replaced by #366

PR #366, opened 2020-01-30
[v4r0] Move travis to github actions and introduce extjs esLint plugin

zmathe:
BEGINRELEASENOTES

CHANGE: Use github actions and introduce extjs eslint plugin

ENDRELEASENOTES


.github/workflows/codestyle.yml
```diff
@@ -0,0 +1,32 @@
...
+          npm install @sencha_eslint-plugin-extjs.7.0.0.tgz --save-dev
+          node_modules/.bin/eslint WebApp
+  pyLint:
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v1
+    - name: Prepare environment
+      run: |
+        conda env create --name webapp-testing environment.yml
```
fstagni: Last change: put a python version matrix at the beginning of this file (with, right now, only python 2.7.13), and use such version in this line (and then remove the python version from the environment.yml)

zmathe: done

PR #370, opened 2020-03-06
adapt to DIRAC PR 4491

fstagni:


fstagni:
This PR won't work in DIRAC v6r22. Indeed, it can only work if and only if DIRAC version is at least v7r1+. In that case I am wondering if we should create another branch...

atsareg:
The PR can target v4r1 then

fstagni:
PR https://github.com/DIRACGrid/DIRAC/pull/4491 is targeting `rel-v7r1`. But v7r1 uses WebApp 4.0. Since the DIRAC PR above effectively changes the APIs then I can't simply target `rel-v4r1`

atsareg:
Then v6r22 and v7r0 will stop at the current patch of v4r0 and v7r1 will use the later patch. In the worst case that some v4r0 fixes will be needed for v6r22, then we will create a new branch.

fstagni:
I don't think that's the best solution, bug fixes might be needed at some point in time for versions used by v6r22 or v7r0. I would prefer do things properly. One possibility is that I make use of the `deprecated` decorator in https://github.com/DIRACGrid/DIRAC/pull/4491, but this might mean duplicating quite some code.

PR #371, opened 2020-03-10
[v4r0] Added StorageOccupancy to accounting web app

fstagni:
BEGINRELEASENOTES

NEW: Added StorageOccupancy to accounting web app

ENDRELEASENOTES


fstagni:
This PR won't work in DIRAC v6r22. Indeed, it can only work if and only if DIRAC version is at least v7r0p15. In that case I am wondering if we should create another branch...

Issue #377, opened 2020-04-03
TransformationClient.getTransformationInputDataQuery no longer exists in DIRAC v7r0


andresailer:
https://github.com/DIRACGrid/WebAppDIRAC/blob/c4a9bae8c1c6d995ed9b22845901ff1230d0a7b9/WebApp/handler/TransformationMonitorHandler.py#L276

As long as webapp v4r0 is used both for v6r22 and v7r0 one cannot simply change the function.

PS: I hotfixed on lbcertifdirac70

fstagni:
How did you hotfix?

fstagni:
I see it should be "replaced" by `transClient.getTransformationMetaQuery(prodid, 'Input')` but this means that the hack would be simply go for a `try..except`. Is this the best solution?

andresailer:
Sorry, the bug report was a bit short.

Probably the try/except like check is the best solution.
Looking for  "Unknown method getTransformationInputDataQuery" in the res['Message'] and then trying `getTransformationMetaQuery`? Or rather first getTransformationMetaQuery and the fallback to the old function.

fstagni:
I think the second option: try the new one, fallback to the old. I can take care later, unless you want to do it.

fstagni:
Fixed.

PR #378, opened 2020-04-03
[v4r0] try..except for TS method change

fstagni:
For #377 

BEGINRELEASENOTES

FIX: try..except for TS method change

ENDRELEASENOTES

### DIFF
--- a/WebApp/handler/TransformationMonitorHandler.py
+++ b/WebApp/handler/TransformationMonitorHandler.py
@@ -273,7 +273,13 @@ class TransformationMonitorHandler(WebHandler):
   def __dataQuery(self, prodid):
     callback = {}
     tsClient = TransformationClient()
-    res = tsClient.getTransformationInputDataQuery(prodid)
+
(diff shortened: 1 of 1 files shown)


WebApp/handler/TransformationMonitorHandler.py
```diff
@@ -273,14 +276,20 @@ def __fileRetry(self, prodid, mode):
   def __dataQuery(self, prodid):
     callback = {}
     tsClient = TransformationClient()
-    res = tsClient.getTransformationInputDataQuery(prodid)
+
+    # FIXME: getTransformationInputDataQuery has been replaced by getTransformationMetaQuery in DIRAC v7r0
+    try:
```
andresailer: There won't be an exception, the Client will just forward any call to the service handler. One has to check the res['Message'] for the Unknown method string

fstagni: Of course! Fixed, thanks.

andresailer (approved):

fstagni:
@atsareg please include this in today's releases

PR #380, opened 2020-04-29
[v4r1] Add option ApplicationList to LaunchPad

TaykYoku:
The `Launchpad` app have predefined sets of values(`Operations/Defaults/Launchpad` in CS). Launchpad handler use it for all VOs, but if some don't want use defaults.. This PR allows to ignore it and use only sets specified in `Operations/<vo>/<setup>/Launchpad/ApplicationList ` option.

BEGINRELEASENOTES

LaunchPad:
NEW: Add option ApplicationList

ENDRELEASENOTES

### DIFF
--- a/WebApp/handler/JobLaunchpadHandler.py
+++ b/WebApp/handler/JobLaunchpadHandler.py
@@ -137,7 +137,7 @@ class JobLaunchpadHandler(WebHandler):
     predefinedSets = {}
     launchpadSections = obj.getSections("Launchpad")
     if launchpadSections['OK']:
-      for section in launchpadSections["Value"]:
+      for section in obj.getValue("Launchpad/ApplicationList", launchpadSections["Value"]):
         predefinedSets[section] = {}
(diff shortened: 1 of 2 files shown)


fstagni:
Please rebase, we should see the `prettier` check succeed.

fstagni:
I don't get why `prettier` fails right now. Can you investigate why, pleas?

TaykYoku:
Yes, I'm just watching. I downgraded my prettier to `1.12.0` version(next version is `2.0.0`), but the error does not disappear. Can we use for testing/formatting last version of the prettier?

fstagni:
Please try with prettier 1.19.1: this is what's installed in GA: https://github.com/DIRACGrid/WebAppDIRAC/blob/integration/.github/workflows/codestyle.yml#L21

TaykYoku:
I think I found the reason

TaykYoku:
done

fstagni (approved):

PR #385, opened 2020-05-26
[v4r0] GA: find the correct DIRAC release to checkout

fstagni:
Also added pylint3k check

### DIFF
--- a/Lib/Compiler.py
+++ b/Lib/Compiler.py
@@ -1,3 +1,7 @@
+from __future__ import print_function
+from __future__ import division
+from __future__ import absolute_import
+
 import tempfile
 import os
 import subprocess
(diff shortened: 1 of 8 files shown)


fstagni:
I'd like to merge this PR and rebase the others. Review OK?

Lib/Compiler.py
```diff
@@ -115,7 +119,7 @@ def __zip(self, staticPath, stack=""):
       if os.path.isfile(zipPath):
         if os.stat(zipPath).st_mtime > os.stat(ePath).st_mtime:
           continue
-      print "%s%s\r" % (n, " " * (20 - len(n))),
+      print("%s%s\r" % (n, " " * (20 - len(n))))
```
andresailer: The comma at the end was to avoid printing a linebreak, together with the `\r` to have some single line counter/bar/?
maybe adding keeps that?
```suggestion
      print("%s%s\r" % (n, " " * (20 - len(n))), end='')
```

fstagni:
Merging, I think it's fine...

PR #388, opened 2020-06-03
fixing the version of prettier

fstagni:


fstagni:
I will merge as soon as GitHub Action finishes successfully.

PR #396, opened 2020-07-17
[v4r0] Fixes for various errors and deprecation warnings

### LINKED ISSUE #387: WebApp is unusable if Google is blocked
If requests to `https://www.google.com/jsapi` are blocked the web app never loads due to the javascript crashing with:

```python
ReferenceError: google is not defined
```

chrisburr:
This:

* removes a bunch of errors and warnings which appear in the developer console when browsing the web app.
* changes the default TLS version in tornado. This is badly needed both for security if not using nginx and as browsers are removing support: https://www.chromestatus.com/feature/5759116003770368
* allows the webapp to be mostly functional if google.com is blocked (fixes #387)

I've also modified the eslint configuration to set the version to ES6. This is needed for 48d7a6d35233556b99cccd7b1a26dd6d40b7a4bc and I think it's reasonable as ES6 is from 2015 and nobody should be using the web app in a browser that is that outdated. If people disagree I'll rebase that commit out of this PR.

I've been using a local copy of the webapp for a couple of weeks against LHCbDIRAC without any issues so these changes should be fairly well tested.

BEGINRELEASENOTES

FIX: WebApp hangs indefinitely if google.com is blocked (#387)
CHANGE: Default to TLS 1.2
FIX: Fix minor deprecation warnings in extjs and modern browsers

ENDRELEASENOTES

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,4 +1,7 @@
 {
+    "parserOptions": {
+        "ecmaVersion": 6,
+    },
     "plugins": [
         "@sencha/extjs"
     ],
--- a/.gitignore
+++ b/.gitignore
@@ -51,3 +51,7 @@ __pycache__
 pytests.xml
 nosetests.xml
 coverage.xml
+
+# Compiliation output
+WebApp/static/extjs/
+*.gz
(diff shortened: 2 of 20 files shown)


TaykYoku:
as for me, it looks fine

fstagni (approved):

Issue #324, opened 2019-06-28
Search feature does not work


chaen:
We noticed that the `ctrl+f` search does not work anymore for the CS and the system administrator (and I suspect the other apps too then).
This is a very desired features :)

andresailer:
It works for the text that is visible on the screen, but not for anything outside of what is visible

zmathe:
It is fine in other applications for example JobMonitor

zmathe:
It is introduced in Extjs 6: https://stackoverflow.com/questions/52513580/sencha-extjs6-0-ext-window-window-diagnostic-window-not-able-to-search-text-l

fstagni:
Should be fixed.

PR #419, opened 2020-11-06
[v4r0] Disable bufferedRenderer in RegistryManager to allow better searching 

chrisburr:
As reported on the forum. I agree it's annoying and I'm not convinced there is any performance benefit for most of DIRAC's pages.

BEGINRELEASENOTES

FIX:  Disable bufferedRenderer in RegistryManager to allow better searching 

ENDRELEASENOTES

@TaykYoku Could you take a look for if there are other places that would benefit from this? (such as the Configuration Manager, though maybe that's too big when fully expanded?)


chaen:
Isn't that related https://github.com/DIRACGrid/WebAppDIRAC/issues/324?

chrisburr:
Yes it is!

TaykYoku:
Probably need to write a "search" method to use buffering..

PR #424, opened 2020-11-19
[v4r2] move web.cfg to main CS

### LINKED ISSUE #368: Make web.cfg optional, move its content to dirac.cfg
Currently, `web.cfg` is always lookedup and needed, otherwise weird stuff happens.  Basically what needs to be done is to use the standard way of dirac:  any `.cfg` on the command line should be treated as a configuration file, otherwise, it looks for the default. 
This allows to put all the Web configuration in the standard `etc/dirac.cfg` and also people wanting to stick with a separate `web.cfg` can just add it in the `run` file of the WebApp

TaykYoku:
BEGINRELEASENOTES
Use web.cfg as example and /WebApp section in dirac.cfg to configure Web portal. Directly depends on [#4829](https://github.com/DIRACGrid/DIRAC/pull/4829).

CHANGE: don't read web.cfg

ENDRELEASENOTES

### DIFF
--- a/Core/App.py
+++ b/Core/App.py
@@ -42,81 +42,6 @@ class App(object):
   def __reloadAppCB(self):
     gLogger.notice("\n !!!!!! Reloading web app...\n")
 
-  def _loadWebAppCFGFiles(self):
-    """
-    Load WebApp/web.cfg definitions
-    """
-    exts = []
-    for ext in CSGlobals.getCSExtensions():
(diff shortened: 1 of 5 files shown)


fstagni:
The PR is IMHO fine, at least code-wise. We need documentation and update of https://github.com/DIRACGrid/DIRAC/wiki/DIRAC-v7r2

PR #431, opened 2021-02-09
Continue loading even if one handler is broken

chaen:
Needs https://github.com/DIRACGrid/DIRAC/pull/4957
Without this, a single broken module is enough to break the whole portal.

BEGINRELEASENOTES
CHANGE: Do not crash at load time because of one buggy handler

ENDRELEASENOTES


TaykYoku:
please, rebase PR, to resolve **Basic tests / eslint (pull_request)**

fstagni (approved):

Issue #252, opened 2016-08-15
SystemAdministrator should directly query ComponentMonitoring


andresailer:
At the moment the SystemAdministrator is querying each server which then might turn to the ComponentMonitoring itself.
https://github.com/DIRACGrid/WebAppDIRAC/blob/integration/WebApp/handler/SystemAdministrationHandler.py#L56

Wouldn't it make sense to first query the ComponentMonitoring directly?

zmathe:
This was the original goal, but it is not implemented that way. It needs further improvement. But this is still faster than browsing the machine file system.

zmathe:
The host are retrieved for the ComponentMonitoring, the host info are retrieved directly from the host. So, the SystemAdministrator page contains information, which are not in the component database for example host info. I would say when the elastic search based component monitoring contains this information, then the SystemAdministrator can be changed...

fstagni:
Looks like fixed.

PR #440, opened 2021-03-03
[v4r2] optimize sysadmin getSysInfo to use directly query ComponentMonitoring

TaykYoku:
BEGINRELEASENOTES
This PR about #252 
Depended from https://github.com/DIRACGrid/DIRAC/pull/5012

CHANGE: optimize sysadmin getSysInfo method to use directly query ComponentMonitoring

ENDRELEASENOTES

### DIFF
--- a/WebApp/handler/SystemAdministrationHandler.py
+++ b/WebApp/handler/SystemAdministrationHandler.py
@@ -23,7 +23,9 @@ class SystemAdministrationHandler(WebHandler):
   def web_getSysInfo(self):
     """ Provide information about hosts state from database
     """
-    client = ComponentMonitoringClient(delegatedDN=self.getUserDN(), delegatedGroup=self.getUserGroup())
+    DN = self.getUserDN()
+    group = self.getUserGroup()
+    client = ComponentMonitoringClient(delegatedDN=DN, delegatedGroup=group)
     result = yield self.threadTask(client.getLogs)
     if not result['OK'] or not len(result['Value']) > 0:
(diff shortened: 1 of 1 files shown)


WebApp/handler/SystemAdministrationHandler.py
```diff
@@ -22,83 +21,41 @@ class SystemAdministrationHandler(WebHandler):
...
     callback = list()
 
     if "hostname" not in self.request.arguments and not self.request.arguments["hostname"][0]:
       self.finish({"success": "false", "error": "Name of the host is absent"})
       return
 
     host = self.request.arguments["hostname"][0]
-    client = SystemAdministratorClient(host, None, delegatedDN=DN,
-                                       delegatedGroup=group)
+    client = SystemAdministratorClient(host, None, delegatedDN=self.getUserDN(),
```
atsareg: I think this does not work as the self.getUserDN() will be resolved at the moment of the code loading.
```suggestion
    userDN = self.getUserDN()
    client = SystemAdministratorClient(host, None, delegatedDN=userDN,
```
and the same for the group

TaykYoku: returned as it was before

atsareg (approved):

WebApp/handler/SystemAdministrationHandler.py
```diff
@@ -89,16 +51,15 @@ def web_getHostData(self):
     """
     DN = self.getUserDN()
     group = self.getUserGroup()
-
+
```
atsareg: remove white spaces here
```suggestion
```

Issue #428, opened 2021-01-19
Site Summary plots are not correct for transfers


phicharp:
In the Site Summary application, when selecting a site, 6 plots are produced: 4 on jobs and 2 on transfers...  It seems plots for jobs do concern the selected site while transfer plots are for all channels... They should only refer to the SEs at the selected site...

phicharp:
In addition, the status of CEs and SEs only have 2 columns: Name and Type... and the interesting information is missing, i.e. AccessType and Status (Active/banned)

fstagni:
@phicharp is it still the case? This version should be in production in LHCb.

phicharp:
Yes, nothing has changed: DMS plots are still for all channels and SE/CE summary have 2 columns only

phicharp:
![Uploading 210323-SiteSummary.png…]()

TaykYoku:
>In addition, the status of CEs and SEs only have 2 columns: Name and Type... and the interesting information is missing, i.e. AccessType and Status (Active/banned)

The "overview" option on the item provide additional information about element. The nuance is that when the "Overview" right panel opens, there is not much space left in a tables, I'm working on it. 

> In the Site Summary application, when selecting a site, 6 plots are produced: 4 on jobs and 2 on transfers... It seems plots for jobs do concern the selected site while transfer plots are for all channels... They should only refer to the SEs at the selected site...

This should be fixed with #444.

TaykYoku:
I think it can be closed and reopen if additional questions arise

PR #449, opened 2021-03-23
[v4r0] add columns to SiteSummary overview tables

TaykYoku:
BEGINRELEASENOTES
Addressing the issues discussed in the #428.

NEW: Add columns to SiteSummary overview tables. 

ENDRELEASENOTES

### DIFF
--- a/WebApp/handler/SiteSummaryHandler.py
+++ b/WebApp/handler/SiteSummaryHandler.py
@@ -158,7 +158,7 @@ class SiteSummaryHandler(ResourceSummaryHandler):
                                          None)
 
       for sestatus in sestatuses['Value']:
-        storageElementsStatus.append([sestatus[0], sestatus[2], sestatus[6]])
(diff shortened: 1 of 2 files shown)


TaykYoku:
@phicharp please correct me if this is what you expected.

New columns added:
 - StatusType
 - Status

Знімок екрана 2021-03-23 о 16 00 41


Is that enough for a overview?

Also note that the first column has a color indicator of the general status of the element.

phicharp:
Yes, this is what I was expecting! Clearly the first 2 columns only was not informative... I have noticed the color code matches the "Status" but this is nice...

phicharp:
Did you also manage to fix the Data management plots for considering only the SEs connected with the site?

TaykYoku:
> Did you also manage to fix the Data management plots for considering only the SEs connected with the site?

Yes, it should already be working properly

PR #451, opened 2021-03-23
[v4r3] fix System, Module --> DIRACSystem, DIRACModule

TaykYoku:
BEGINRELEASENOTES
This PR fix v7.3.a0 hackathon issues.

FIX: fix System, Module --> DIRACSystem, DIRACModule

ENDRELEASENOTES


fstagni:
The change actually happened in DIRAC v7r2: https://github.com/DIRACGrid/DIRAC/pull/4956 which means that you should target v4r2, not v4r3.

TaykYoku:
Moved to #456

PR #453, opened 2021-03-25
[v4r0] fix plot update

### LINKED ISSUE #452: [v4r0] Presenter: "apply" button not working
For a little while, the "apply" button of the presenter does not work properly. The small image is updated, but when you click on it to get the big picture, it is still the old one. I suspect strongly that it comes from that  https://github.com/DIRACGrid/WebAppDIRAC/pull/437/files

Can this be urgently fixed please ? 

How to reproduce (Firefox):

Create a plot, for example
image

Now, change the selector (here `Final Transfer Status`), and click `Apply`

image

All nice, but if you click on the image to have it big, you get the old one

image

TaykYoku:
BEGINRELEASENOTES
This PR resolve issue #452.

FIX: use setSrc to replaceImage

ENDRELEASENOTES

### DIFF
--- a/WebApp/static/core/js/utils/PlotView.js
+++ b/WebApp/static/core/js/utils/PlotView.js
@@ -529,6 +529,7 @@ Ext.define('Ext.dirac.utils.PlotView', {
                         });
                   },
 
+                  // When new plot added on the wall generate image
                   added : function(container, pos, eOpts) {
                     var me = this;
                     me.setLoading(true);
--- a/WebApp/static/core/js/utils/Presenter.js
+++ b/WebApp/static/core/js/utils/Presenter.js
@@ -359,11 +359,12 @@ Ext.define('Ext.dirac.utils.Presenter', {
         var me = this;
         return me.lastClickedImage;
       },
+      // When you apply new input data for example, need to update image
       replaceImage : function(oldImage, newImage) {
         var me = this;
         var img = me.getComponent(oldImage.id);
         img.plotParams = newImage['params'];
-        img.updateSrc(newImage['src']);
+        img.setSrc(newImage['src']);
       },
       setColumnWidth : function(column) {
         var me = this;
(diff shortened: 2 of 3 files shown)


chaen:
Out of curiosity, can you explain what is the difference between setSrc and updateSrc, and why we have both ?

TaykYoku:
In a nutshell, the `setSrc` method set/change the `src` value of the `Ext.Img` object. The `updateSrc` method update only the `src` attribute from HTML dom element using `setAttribute` method. That is, `updateSrc` updated the image in the page html element, but did not change the link in the image object. I hope this is the answer to your question. Here are links to sources with some comments for more immersion:
[setSrc doc](https://docs.sencha.com/extjs/6.2.0/modern/Ext.Img.html#method-setSrc)
[makeSetter source](https://docs.sencha.com/extjs/6.2.0/classic/src/Config.js.html)
[updateSrc doc](https://docs.sencha.com/extjs/6.2.0/modern/Ext.Img.html#method-updateSrc)
[updateSrc source](https://docs.sencha.com/extjs/6.2.0/modern/src/Img.js.html)

chaen:
I see thanks ! But in that case shouldn't we always use `setSrc`

TaykYoku:
`setSrc` set new src to image object and update content.
Знімок екрана 2021-04-08 о 00 23 47
When we open it we see source image:
Знімок екрана 2021-04-08 о 00 24 30
Then `updateSrc` update DOM object from other source url
Знімок екрана 2021-04-08 о 00 25 27
But when we open it we see image from old source url.
Знімок екрана 2021-04-08 о 00 25 55
If we `setSrc` this new image generator url, we can see that image updated in DOM
Знімок екрана 2021-04-08 о 00 26 51
When open it we see updated image from this new url.
Знімок екрана 2021-04-08 о 00 27 46
But if we try to `setSrc` with the same source URL we can see that DOM object not updated
Знімок екрана 2021-04-08 о 00 43 16

So I modify `replaceImage` to choose between `setSrc` and `updateSrc`. #460

PR #461, opened 2021-04-14
prepare for pip install

TaykYoku:
BEGINRELEASENOTES
Preparing a project for installation using `pip install`.

CHANGE: move main source code to src
NEW: add pyproject.toml, setup.py
CHANGE: add project description to setup.cfg
NEW: move tests things to tests
CHANGE: rename README.md to README.rst

ENDRELEASENOTES


setup.cfg
```diff
@@ -1,4 +1,62 @@
...
+
+[options.packages.find]
+where=src
+
+[options.extras_require]
+server =
+testing =
+
+[options.entry_points]
+webappdirac =
```
chrisburr:
```suggestion
dirac =
```

setup.cfg
```diff
@@ -1,4 +1,62 @@
-# This file should be used as the pep8/pycodestyle and autopep8 config
+###############################################################################
+# Setuptools
+###############################################################################
+
+[metadata]
+name = WebAppDIRAC
+description = WebAppDIRAC is a prortal for the DIRAC software.
```
chrisburr:
```suggestion
description = WebAppDIRAC is a portal for the DIRAC software.
```

setup.cfg
```diff
@@ -1,4 +1,62 @@
...
+    #autopep8 ==1.3.3
+    coverage
+    requests >=2.9.1
+    six
+    #mock >=1.0.1
+    #pylint >=1.6.5
+    #pytest >=3.6
+    #pytest-cov >=2.2.0
+    #pytest-mock
+    #tornado >=5.0.0
```
chrisburr: I'm not sure if this is correct but you'll need to include the package data:
```suggestion

zip_safe = False
include_package_data = True

[options.package_data]
* =  WebApp/static
```

You can check by running `python setup.py sdist bdist_wheel` and unzipping the files which are generated to make sure everything you expect is included.

pyproject.toml
```diff
@@ -0,0 +1,10 @@
+[build-system]
+requires = ["setuptools>=42", "wheel", "setuptools_scm[toml]>=3.4"]
+build-backend = "setuptools.build_meta"
+
+# Enable setuptools_scm to compute the version number from the most recent tag
+# https://github.com/pypa/setuptools_scm/#pyprojecttoml-usage
```
chrisburr, line 6: Is there a reason for disabling `setuptools_scm`?

TaykYoku: not at all, but had to test how it works. `setuptools_scm` collects the version from git which is not supported in pypi therefore it was added the test version in setup  manual

TaykYoku: in theory I should create a tag with the correct version and uncomment it

src/WebAppDIRAC/scripts/dirac_webapp_run.py
```diff
@@ -5,15 +5,17 @@
...
 from DIRAC.FrameworkSystem.Client.Logger import gLogger
 
 from WebAppDIRAC.Core.App import App
 
 __RCSID__ = "$Id$"
 
-if __name__ == "__main__":
 
+@DiracScript()
+def main()
```
chrisburr:
```suggestion
def main():
```

setup.cfg
TaykYoku: tested with `include_package_data`, with `options.package_data ` and without, the result is no different. in all cases js, tpl and others are available in the archive

pyproject.toml
TaykYoku: fixed

setup.cfg
```diff
@@ -1,4 +1,66 @@
...
+    Topic :: System :: Distributed Computing
+
+[options]
+python_requires = >=3.8
+package_dir=
+    =src
+packages = find:
+
+install_requires =
+    dirac
```
chrisburr: I guess this should have a version associated with it as I guess this tends to be tightly coupled to DIRAC?
```suggestion
    dirac ~=7.3.0a0
```

setup.cfg
```diff
@@ -1,4 +1,66 @@
...
+    diraccfg
+    autopep8 ==1.3.3
+    coverage
+    requests >=2.9.1
+    six
+    mock >=1.0.1
+    pylint >=1.6.5
+    pytest >=3.6
+    pytest-cov >=2.2.0
+    pytest-mock
```
chrisburr: Many of these are test dependencies and `tornado` should be added as it's used directly. I don't think there is anything else but please check:
```suggestion
    requests >=2.9.1
    tornado
    six
```

setup.cfg
```diff
@@ -1,4 +1,66 @@
...
+include_package_data = True
+
+[options.package_data]
+* = WebApp/static
+
+[options.packages.find]
+where=src
+
+[options.extras_require]
+testing =
```
chrisburr, line 45: Test only dependencies can go here

tests/runPylint.sh
```diff
@@ -1,7 +1,7 @@
 #!/bin/bash
 
 if [[ "${CHECK}" == "pylintPY3K"  ]]; then
-  find WebApp Lib Core scripts -name "*.py" -and -not -name 'pep8_*' -exec pylint -E --rcfile=.pylintrc3k --py3k --msg-template="{path}:{line}: [{msg_id}({symbol}), {obj}] {msg}" {} +
+  find src/WebAppDIRAC/* -name "*.py" -and -not -name 'pep8_*' -exec pylint -E --rcfile=tests/.pylintrc3k --py3k --msg-template="{path}:{line}: [{msg_id}({symbol}), {obj}] {msg}" {} +
```
chrisburr: No need for the `*`:
```suggestion
  find src/WebAppDIRAC/ -name "*.py" -and -not -name 'pep8_*' -exec pylint -E --rcfile=tests/.pylintrc3k --py3k --msg-template="{path}:{line}: [{msg_id}({symbol}), {obj}] {msg}" {} +
```

tests/runPylint.sh
```diff
@@ -1,7 +1,7 @@
...
 else
-  find WebApp Lib Core scripts -name "*.py" -and -not -name 'pep8_*' -exec pylint -E --rcfile=.pylintrc --msg-template="{path}:{line}: [{msg_id}({symbol}), {obj}] {msg}" {} +
+  find src/WebAppDIRAC/* -name "*.py" -and -not -name 'pep8_*' -exec pylint -E --rcfile=.pylintrc --msg-template="{path}:{line}: [{msg_id}({symbol}), {obj}] {msg}" {} +
```
chrisburr: No need for the `*`:
```suggestion
  find src/WebAppDIRAC/ -name "*.py" -and -not -name 'pep8_*' -exec pylint -E --rcfile=.pylintrc --msg-template="{path}:{line}: [{msg_id}({symbol}), {obj}] {msg}" {} +
```

setup.cfg
TaykYoku: Correct me, IIUC then WebAppDIRAC should be installed with the dependence of the `dirac` package and it has the dependence of `tornado`, do it still needed to specify both there and here?

chrisburr: You should depend on whatever WebAppDIRAC uses directly. The way to think about it is "if DIRAC suddenly stopped using tornado would WebAppDIRAC still need it?" and in this case the answer is yes.

TaykYoku: Thanks, now I understand what you mean

src/WebAppDIRAC/WebApp/static/core/js/utils/canvg-1.3/StackBlur.js
```diff
@@ -1,611 +1,611 @@
...
-}
-
-function BlurStack()
-{
-	this.r = 0;
-	this.g = 0;
-	this.b = 0;
-	this.a = 0;
-	this.next = null;
+/*
```
chrisburr, line 1: What is the reason for changing these files? As they're vendored dependencies we should probably leave them in their upstream state.

TaykYoku: these files have been just moved without any change

chrisburr: They have been changed a lot:

Screenshot 2021-04-22 at 16 38 39

TaykYoku: so I see, but the content is the same .. I'll try to delete and move them again

chrisburr: It had been reformatted, I guess accidentally. I can fix it locally using this but I don't have permission to push:
```bash
$ git checkout upstream/integration WebApp/static/core/js/utils/canvg-1.3/
Updated 4 paths from 108e6fa
$ rm -rf src/WebAppDIRAC/WebApp/static/core/js/utils/canvg-1.3/
$ git mv WebApp/static/core/js/utils/canvg-1.3/ src/WebAppDIRAC/WebApp/static/core/js/utils/canvg-1.3/
$ git commit -m "Restore content of canvg-1.3"
[v4r3_pyPA3 9dc9af1] Restore content of canvg-1.3
 4 files changed, 3759 insertions(+), 3759 deletions(-)
```

TaykYoku: as far as i understood my git when moving this files removes the CR from the file encoding, I don't know how typical this behavior is, but I'll try to somehow return the CR to keep the originality of the files..

chrisburr: Yeah file encoding and line endings can be a real pain...

It's fixed in https://github.com/DIRACGrid/WebAppDIRAC/pull/467.

PR #473, opened 2021-05-26
[v4r0] Use listStatesForWeb endpoint

chrisburr:
Requires https://github.com/DIRACGrid/DIRAC/pull/5159

BEGINRELEASENOTES

NEW: Use listStatesForWeb endpoint to make the Public State Manager faster

ENDRELEASENOTES


TaykYoku:
restart tests

PR #481, opened 2021-06-18
Various Python 3 fixes

chrisburr:
BEGINRELEASENOTES

FIX: Use a PEP-440 style version string if running with Python 3
FIX: Miscellaneous Python 3 compatibility fixes

ENDRELEASENOTES

### DIFF
--- a/src/WebAppDIRAC/Lib/WebHandler.py
+++ b/src/WebAppDIRAC/Lib/WebHandler.py
@@ -34,26 +34,13 @@ class WErr(tornado.web.HTTPError):
     super(WErr, self).__init__(code, str(msg) or None)
     for k in kwargs:
       setattr(self, k, kwargs[k])
-    self.ok = False
     self.msg = msg
     self.kwargs = kwargs
-
-  def __str__(self):
-    return super(WErr, self).__str__()
 
   @classmethod
   def fromSERROR(cls, result):
     """ Prevent major problem with % in the message """
     return cls(500, result['Message'].replace("%", ""))
-
-
-class WOK(object):
-
-  def __init__(self, data=False, **kwargs):
-    for k in kwargs:
-      setattr(self, k, kwargs[k])
-    self.ok = True
-    self.data = data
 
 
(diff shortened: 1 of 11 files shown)


src/WebAppDIRAC/Lib/WebHandler.py
```diff
@@ -357,7 +357,7 @@ def __checkPath(self, setup, group, route):
 
   def get(self, setup, group, route):
     if not self._pathResult.ok:
-      raise self._pathResult
+      raise Exception(self._pathResult)
```
chrisburr: I'm less sure about these two changes as I can't figure out what catches the exceptions.

@TaykYoku Any ideas?

TaykYoku: In this place self._pathResult  is `WErr` that child of the `tornado.web.HTTPError`. `Exception(self._pathResult)` will also work but the output will change.
What is the problem with using WErr?

chrisburr: Ah right thanks, I understand the actual issue now. The issue was a `pylint` saying:
```python
************* Module src.WebAppDIRAC.Lib.WebHandler
src/WebAppDIRAC/Lib/WebHandler.py:357:6: E0710: Raising a new style class which doesn't inherit from BaseException (raising-non-exception)
src/WebAppDIRAC/Lib/WebHandler.py:397:6: E0710: Raising a new style class which doesn't inherit from BaseException (raising-non-exception)
```
It's debatable if this is a false-positive  or not in the general case but for DIRAC it definitely is a false-positive.

That said it seems that `WOK` is basically unused so I think it can just be removed. Does 0fe8bc7 look reasonable to you?

TaykYoku: Yes, 0fe8bc7 looks good for me

TaykYoku:
Is it ready to merge?

chrisburr:
Yup, go ahead 👍

PR #482, opened 2021-06-21
[integration] Python 3 compatibility fixes

chrisburr:
Mostly fixes bytes/str mix-ups in the code which are hard to spot without running it. Hopefully this PR systematically finds them all though it's big enough that I'm bound to have made mistakes.

There is a lot of potential for clean up and simplifying the code however I don't want to make this PR any bigger than it already is.

BEGINRELEASENOTES

FIX: Mostly fixes bytes/str mix-ups in the code which are hard to spot without running it; optimization
NEW: add git Python 3 tests

ENDRELEASENOTES

### DIFF
--- a/src/WebAppDIRAC/WebApp/handler/ProxyManagerHandler.py
+++ b/src/WebAppDIRAC/WebApp/handler/ProxyManagerHandler.py
@@ -157,7 +157,7 @@ class ProxyManagerHandler(WebHandler):
       req['UserGroup'] = usersgroup
 
     persistent = self.get_argument("persistent", "")
-    if persistent in ["True", "False"]:
+    if usersgroup and persistent in ["True", "False"]:
       req["PersistentFlag"] = persistent
     before = int(self.get_argument("expiredBefore", "0"))
     after = int(self.get_argument("expiredAfter", "0"))
--- a/src/WebAppDIRAC/WebApp/handler/ProxyManagerHandler.py
+++ b/src/WebAppDIRAC/WebApp/handler/ProxyManagerHandler.py
@@ -145,11 +145,11 @@ class ProxyManagerHandler(WebHandler):
     start = 0
     limit = 25
 
-    if "limit" in self.request.arguments and len(self.request.arguments["limit"][0]) > 0:
-      limit = int(self.request.arguments["limit"][0])
+    if "limit" in self.request.arguments and len(self.request.get_argument("limit")) > 0:
+      limit = int(self.request.get_argument("limit"))
 
-    if "start" in self.request.arguments and len(self.request.arguments["start"][0]) > 0:
-      start = int(self.request.arguments["start"][0])
+    if "start" in self.request.arguments and len(self.request.get_argument("start")) > 0:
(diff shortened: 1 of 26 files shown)


src/WebAppDIRAC/WebApp/handler/UPHandler.py
```diff
@@ -301,15 +265,8 @@ def web_listPublicStates(self):
...
-    try:
-      name = self.request.arguments['name'][-1]
-    except KeyError as excp:
-      raise WErr(400, "Missing %s" % excp)
-    try:
-      access = self.request.arguments['access'][-1].upper()
-    except KeyError as excp:
-      access = 'ALL'
-
+    name = self.get_argument("name")
```
TaykYoku, line 268: Why do not raise `WErr` here when argument is absent?

chrisburr: `get_argument` raises a `MissingArgumentError` which is a subclass of `HTTPError` so I think there is no need to explicitly raise a `WErr`?

* https://www.tornadoweb.org/en/branch5.1/web.html#tornado.web.RequestHandler.get_argument

src/WebAppDIRAC/WebApp/handler/ProxyManagerHandler.py
```diff
@@ -108,93 +107,65 @@ def web_deleteProxies(self):
...
+    users = list(json.loads(self.get_argument("username", "[]")))
+    if users:
+      req['UserName'] = users
+
+    usersgroup = list(json.loads(self.get_argument("usergroup", "[]")))
+    if usersgroup:
+      req['UserGroup'] = usersgroup
+
+    persistent = self.get_argument("persistent", "")
+    if persistent in ["True", "False"]:
```
TaykYoku: Is there a reason to change the logic here?
```suggestion
    if usersgroup and persistent in ["True", "False"]:
```

TaykYoku (approved):
Looking at it, it looks like a good job

src/WebAppDIRAC/WebApp/handler/ProxyManagerHandler.py
chrisburr: Well spotted! I hadn't noticed the structure was inconsistent with other similar lines.

I'm not sure if this was a bug as the original code used both `"usergroup"` and `"usersgroup"` and `"usersgroup"` doesn't appear anywhere else in the code:

```python
    if "usergroup" in self.request.arguments:
      usersgroup = list(json.loads(self.request.arguments['usergroup'][-1]))
      if len(usersgroup) > 0:
        req['UserGroup'] = usersgroup

    if "usersgroup" in self.request.arguments and len(self.request.arguments["persistent"]) > 0:
      if str(self.request.arguments["persistent"]) in ["True", "False"]:
        req["PersistentFlag"] = str(self.request.arguments["persistent"])
```

and most lines like this looked like below with the variable name being listed three times:

```python
    if "start" in self.request.arguments and len(self.request.arguments["start"][0]) > 0:
      start = int(self.request.arguments["start"][0])
```

PR #484, opened 2021-06-22
[integration] More Python 3 fixes

### LINKED ISSUE #485: Distributing the WebApp for Python 3 installations
Fundamentally three are two things which are specific to distributing the WebApp (and extensions thereof) that need to be handled:

* There are javascript sources that need to be compiled
* Some ExtJS resources need to be installed (images/fonts/themes/more JS)
* nginx needs to be able to find and serve the static

For Python 2 installations this is currently handled by:

1. The WebApp compiler command is called as part of the release process which causes everything to be included in the tarball that is uploaded
2. Extensions do exactly the same thing
3. nginx can find the files at `/opt/dirac/pro/WebAppDIRAC/WebApp/static/`

For the `pip install`ed Python 3 based installations these can be handled as:

### Part 1: Compilation

The setuptools build process is able to run arbitrary code. The simplest approach here would be run the compiler in the docker container by modifying the `setup.py` to be something like:

```python
from setuptools import Command, setup
# Note: distutils must be imported after setuptools
from distutils import log
from setuptools.command.develop import develop as _develop
from wheel.bdist_wheel import bdist_wheel as _bdist_wheel


class build_extjs_sources(Command):
    user_options = []

    def initialize_options(self):
        pass

    def finalize_options(self):
        pass

    def get_inputs(self):
        return []

    def get_outputs(self):
        return []

    def run(self):
        path = os.path.abspath(os.getcwd())
        cmd = [
            "docker",
            "run",
            "--rm",
            f"-v={path}:/shared",
            "-w=/shared",
            "diracgrid/dirac-distribution",
            "/dirac-webapp-compile.py",
            "-D=/shared/src",
            "-n=WebAppDIRAC",
        ]
        log.info('> %r', cmd)
        subprocess.check_call(cmd)


class develop(_develop):
    def run(self):
        self.run_command("build_extjs_sources")
        super().run()


class bdist_wheel(_bdist_wheel):
    def run(self):
        self.run_command("build_extjs_sources")
        super().run()


cmdclass = {
    "develop": develop,
    "bdist_wheel": bdist_wheel,
    "build_extjs_sources": build_extjs_sources,
}

setup(cmdclass=cmdclass)
```

For releases this will run in GitHub actions as part of making the "wheel" to upload to PyPI. And anyone will then be able to install the release without having docker installed.

For the installation of development versions, e.g. `pip install git+https://.../WebAppDIRAC.git@integration` the `setup.py` will be executed causing `docker` to be ran. Alternatively singularity would be slightly lighter or the WebApp compiler script could be integrated with the above `build_extjs_sources.run` function (see part 2).

One thing to consider is that the wheels produced for WebApp DIRAC are relatively large (~41MB) and PyPI has a 2GB cap (that can be increased, [details here](https://github.com/pypa/pypi-support#guidelines-for-upload-limit-requests)). The main cause is the `static/extjs/` directory which is just copied from the ExtJS SDK and isn't changed between versions. I wonder if the best solution would be to make a `DIRACWebAppResources` package which can be uploaded only when the ExtJS SDK is changed. This then makes WebAppDIRAC ~5MB. Thoughts?

### Part 2: Extensions

WebApp extensions will also need to include the above code in their `setup.py` which is probably burdensome. Instead there could be another new package, similar to how [Jupyter handles this](https://github.com/jupyter/jupyter-packaging). This would simplify extension's `setup.py` to be something like:

```python
from setuptools import setup
from diracweb_packaging import extjs_builder_cmdclass

setup(cmdclass=extjs_builder_cmdclass)
```

As well as adding this to their `pyproject.toml`:

```ini
[build-system]
requires = ["diracweb-packaging~=1.0"]
```

This might also lend itself towards moving the `WebAppCompiler` to this new package and removing the dependency on `docker`. At the same time it can be made to do incremental recompiles to make it less painful to develop the WebApp, i.e. `pip install -e .` would recompile what is changed and nothing else and it would just-work. You could also do `python setup.py build_extjs_sources` to achieve the same thing. To improve the portability of this it could have three ways of running the sencha commands, from `$PATH`, using `singularity`, using `docker`.

### Part 3: nginx

Currently the nginx configuration relies on the static directories being fixed due to the `pro` link whereas for Python 3 installations this won't exist in the same way (see https://github.com/DIRACGrid/DIRAC/pull/5215). Instead the static resources will be in `versions/vX.Y.Z-XXXXXXXXXX-x86_64/lib/pythonX.Y/site-packages/WebAppDIRAC/WebApp/static/` and this path will change for every release.

I'm not sure how to handle this. Perhaps the best thing that can be done is to add a new folder, `/opt/dirac/webRoot/resources` which will contain symlinks like:

```
WebAppDIRAC -> /opt/dirac/versions/vX.Y.Z-XXXXXXXXXX-x86_64/lib/pythonX.Y/site-packages/WebAppDIRAC/WebApp/static
LHCbWebDIRAC -> /opt/dirac/versions/vX.Y.Z-XXXXXXXXXX-x86_64/lib/pythonX.Y/site-packages/LHCbWebDIRAC/WebApp/static
```

When WebAppDIRAC is launched in a server installation and `DevelopMode` is off, it will make sure these symlinks correspond to the current installation.

## Summary of proposed actions

* Create a `DIRACWebAppResources` package
* Create a `diracweb_packaging` package
* Make `WebAppDIRAC` update the symlinks in `/opt/dirac/webRoot/resources` each time it's launched

chrisburr:
Closes #485


chrisburr:
This is ready for review but will needs several other things before the CI passes:
* https://github.com/DIRACGrid/management/pull/41
* https://github.com/DIRACGrid/management/pull/42
* https://github.com/DIRACGrid/DIRAC/pull/5230
* Upload `DIRACWebAppResources` to PyPI after copying to the DIRACGrid organisation: https://github.com/chrisburr/DIRACWebAppResources
* Upload `dirac-webapp-packaging` to PyPI after copying to the DIRACGrid organisation and removing the hack which uses my account on dockerhub: https://github.com/chrisburr/dirac-webapp-packaging

I hacked it to use test-PyPI in the penultimate commit to show it was working.

TaykYoku (commented):
at first glance all is well for me

chrisburr:
@TaykYoku This should be ready for review/merge now

I thought I could avoid the `_createStaticSymlinks` mess but it's actually the least bad option as deploying the static assets elsewhere is hard without either breaking Python 2 installs or duplicating everything.

PR #506, opened 2021-07-28
[v4r2] update datetime field behavior

TaykYoku:
BEGINRELEASENOTES
Update datetime selector.

CHANGE: change datetime field behavior in selector, use Time field

ENDRELEASENOTES


fstagni:
What would be the result?

TaykYoku:
Time field should return the same result:
Hours + ':' + minutes

PR #545, opened 2021-09-17
[v4r4] Support AccountingMonitor permission

sfayer:
Hi,

This patch simply adds the AccountingMonitor to the list of permissions that get the full filter list in the accounting plots (to match https://github.com/DIRACGrid/DIRAC/pull/5392).

I wasn't sure which versions of WebApp are still supported, but perhaps this could also be backported to all other active versions?

Regards,
Simon

BEGINRELEASENOTES
CHANGE: Enable all filter fields for AccountingMonitor role
ENDRELEASENOTES


fstagni:
I think you should port it to v4r1.

TaykYoku:
Resolved with #547.

PR #566, opened 2021-10-07
[v4r3] fix CI

TaykYoku:
BEGINRELEASENOTES

FIX: Update CI according to the compatible DIRAC branch.

ENDRELEASENOTES


setup.cfg
```diff
@@ -40,7 +40,7 @@ where=src
 [options.extras_require]
 server =
     # These dependecies are made optional to make it easier to build extensions
-    dirac ~=7.3.0a0
+    dirac ~=7.3.5
```
fstagni, line 43: Mmm I am no expert here but this does not look correct to me.

.github/workflows/deployment.yml
```diff
@@ -44,13 +42,11 @@ jobs:
             --token="${{ secrets.GITHUB_TOKEN }}" \
             --repo="${{ github.repository }}" \
             --version="v${NEW_STYLE}" \
-            --rev="$(git --git-dir ./WebAppDIRACRepo/.git rev-parse HEAD)" \
-            --release-notes-fn="WebAppDIRACRepo/release.notes"
```
fstagni, line 48: What was this doing, why was it removed?

setup.cfg
TaykYoku: About `~=` operator [here](https://www.python.org/dev/peps/pep-0440/#compatible-release).
[7.3.5](https://github.com/DIRACGrid/DIRAC/releases/tag/v7.3.5) is the latest DIRAC 7.3 patch, so the new 4.3 WebAppDIRAC patches require  DIRAC 7.3.5+ but not DIRAC 8.\*.
WebAppDIRAC 5.0 require DIRAC 8.0.*.

.github/workflows/deployment.yml
TaykYoku: IIUC DIRAC 7.3 does not support `--release-notes-fn` for *make_release*. It comes with DIRAC 8.0.
https://github.com/DIRACGrid/DIRAC/blob/1bbeccf063bfc3db2a7e9b80fb854a51244cd34f/.github/workflows/make_release.py#L50-L56

setup.cfg
TaykYoku: As far as I understand, this is the behavior we expect, no?

PR #573, opened 2021-10-19
[v4r2] Add pre-commit and run it

chrisburr:
BEGINRELEASENOTES

CHANGE: Add pre-commit configuration and run it

ENDRELEASENOTES


chrisburr:
@TaykYoku Can you review this please? The relavent commits are:

* 9c5bfb4
* f740a46
* 55e6ca3
* cf275b3

.github/workflows/codestyle.yml
```diff
@@ -3,15 +3,19 @@ name: Basic tests
 on: [push, pull_request]
 
 jobs:
-  eslint:
+  pre-commit:
     runs-on: ubuntu-latest
+    if: github.event_name != 'push' || github.repository == 'DIRACGrid/DIRAC'
```
TaykYoku: I don't think it will run on `DIRACGrid/DIRAC`
```suggestion
    if: github.event_name != 'push' || github.repository == 'DIRACGrid/WebAppDIRAC'
```

WebApp/static/oidc/oidc-client/dist/oidc-client.min.js
```diff
@@ -1,46 +1,15428 @@
...
-/*!
+var Oidc = (function (t) {
```
TaykYoku: I'm not sure it should be stylized and all the minified versions like as `*.min.js`

TaykYoku:
@chrisburr  Thanks, for this PR. It looks good to me. One thing, I think it's better to skip oidc.

WebApp/static/oidc/oidc-client/dist/oidc-client.min.js
chrisburr: Well spotted and yeah I agree. I've excluded the entire `WebApp/static/oidc` directory from pre-commit:
```yaml
exclude: |
  (?x)^(
      WebApp/static/oidc/|
      WebApp/static/core/pictos/
  ).*$
```

TaykYoku:
@chrisburr What is the next step?
Do I have to merge this with rel-v4r3 first?

chrisburr:
>  Do I have to merge this with rel-v4r3 first?

No I'll make a PR for adding the sweeper so you should never need to merge the branches into each other again.

PR #575, opened 2021-10-19
[5.0] Add pre-commit and run it

chrisburr:
BEGINRELEASENOTES

CHANGE: Add pre-commit configuration and run it

ENDRELEASENOTES


.pre-commit-config.yaml
```diff
@@ -0,0 +1,38 @@
...
+      - id: trailing-whitespace
+      - id: end-of-file-fixer
+      - id: check-yaml
+      - id: check-added-large-files
+
+  - repo: https://github.com/psf/black
+    rev: 21.9b0
+    hooks:
+      - id: black
+        additional_dependencies: [".[python2]"]
```
TaykYoku: Is this option necessary if WebApp 5.0 and DIRAC 8.0 are completely python3?

pyproject.toml
```diff
@@ -7,3 +7,7 @@ build-backend = "setuptools.build_meta"
 [tool.setuptools_scm]
 # Avoid letting setuptools_scm use old style tags (i.e. vXrYpZ)
 git_describe_command = "git describe --dirty --tags --long --match *[0-9]* --exclude v[0-9]r* --exclude v[0-9][0-9]r*"
+
+[tool.black]
+line-length = 120
+target-version = ['py27']
```
TaykYoku: Maybe it's better to use python3

.pre-commit-config.yaml
chrisburr: :+1: Probably not for WebApp (vanilla DIRAC might need it for the pilot stuff but even that is probably okay given we're compatible with both Python 2 and 3 for pilot submission).

TaykYoku:
@chrisburr Thanks, good job!

PR #577, opened 2021-10-20
[v4r3] LHCbDIRAC hackathon fixes

chrisburr:
BEGINRELEASENOTES

FIX: Correct pattern used for handler.PATH_RE to accept urls like `web_t1sites`

ENDRELEASENOTES

### DIFF
--- a/src/WebAppDIRAC/Core/HandlerMgr.py
+++ b/src/WebAppDIRAC/Core/HandlerMgr.py
@@ -112,7 +112,7 @@ class HandlerMgr(object):
             baseRoute = "/%s%s" % (self.__baseURL or "", baseRoute)
             # Set properly the LOCATION after calculating where it is with helpers to add group and setup later
             handler.LOCATION = handlerRoute
-            handler.PATH_RE = re.compile("%s(%s/[A-z0-9_]+|.)" % (baseRoute, handlerRoute))
+            handler.PATH_RE = re.compile("%s(%s/[A-z0-9_-]*)" % (baseRoute, handlerRoute))
             handler.URLSCHEMA = "/%s%%(setup)s%%(group)s%%(location)s/%%(action)s" % (self.__baseURL)
             if issubclass(handler, WebSocketHandler):
                 handler.PATH_RE = re.compile("%s(%s)" % (baseRoute, handlerRoute))
--- a/src/WebAppDIRAC/Core/HandlerMgr.py
+++ b/src/WebAppDIRAC/Core/HandlerMgr.py
@@ -112,7 +112,7 @@ class HandlerMgr(object):
             baseRoute = "/%s%s" % (self.__baseURL or "", baseRoute)
             # Set properly the LOCATION after calculating where it is with helpers to add group and setup later
             handler.LOCATION = handlerRoute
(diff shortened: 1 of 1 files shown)


src/WebAppDIRAC/Core/HandlerMgr.py
```diff
@@ -112,7 +112,7 @@ def __calculateRoutes(self):
             baseRoute = "/%s%s" % (self.__baseURL or "", baseRoute)
             # Set properly the LOCATION after calculating where it is with helpers to add group and setup later
             handler.LOCATION = handlerRoute
-            handler.PATH_RE = re.compile("%s(%s/[A-z]+|.)" % (baseRoute, handlerRoute))
+            handler.PATH_RE = re.compile("%s(%s/[A-z0-9_]+|.)" % (baseRoute, handlerRoute))
```
chrisburr: @TaykYoku I don't understand why the `|.` is here from [9dec695ef156fed211e24b16ca77e9601aa66c69](https://github.com/DIRACGrid/WebAppDIRAC/commit/9dec695ef156fed211e24b16ca77e9601aa66c69#diff-6fe0ffa70f2c7586e88aee70441704bad710fa88f8eab5e57facd297f64e5a9dL108-R119). Can you explain?

TaykYoku: It seems that matching doesn't work without it, but I agree `|.` looks bad. I believe that `re.compile("%s(%s/[A-z0-9_-]*)"` should work correct.
```suggestion
            handler.PATH_RE = re.compile("%s(%s/[A-z0-9_-]*)" % (baseRoute, handlerRoute))
```

TaykYoku: At least I tested it for integration and it seems to work as it should.

Request: `/DIRAC/s:DIRAC-CertifOauth/g:dirac_user/Accounting/getPlotImg`
handler.PATH_RE : `re.compile('/DIRAC(?:/s:([\\w-]*)/g:([\\w.-]*))?(/Accounting/[A-z0-9_]*)')`
match object: `<re.Match object; span=(0, 61), match='/DIRAC/s:DIRAC-CertifOauth/g:dirac_user/Accountin>`
after grouping (setup, group, path): `('DIRAC-CertifOauth', 'dirac_user', '/Accounting/getPlotImg')`

chrisburr: > It seems that matching doesn't work without it

What is "it"?

>  believe that `re.compile("%s(%s/[A-z0-9_-]*)"` should work correct.

I'm pretty sure it's impossible for DIRAC to have a route with a hyphen in it given the names are taken from Python function names?

TaykYoku: > What is "it"?

`|.`

chrisburr: Then I don't understand how "It seems that matching doesn't work without it" can be the case when you've merged this PR?

Also I see yoou changed `+` to `*` in the regex, why?

TaykYoku: Yes, I agree, I was in a hurry to merge this PR. I will make a new one with fixes.

PR #593, opened 2021-10-24
[v4r3] check routing regex

TaykYoku:
BEGINRELEASENOTES
Check and provide correct routing regex. Look #557 discussion.

FIX: provide correct routing regex

ENDRELEASENOTES


chrisburr (approved):

TaykYoku:
@chrisburr  Is there any documentation about DIRACGridBot?

PR #602, opened 2021-11-04
[5.0] Use DIRAC 8.x style deployment for WebAppDIRAC as well

chrisburr:
This modifies the deployment job to be like DIRAC 8.x releases use. See https://dirac.readthedocs.io/en/integration/DeveloperGuide/ReleaseProcedure/index.html for details.

### DIFF
--- a/.github/workflows/deployment.yml
+++ b/.github/workflows/deployment.yml
@@ -10,13 +10,13 @@ on:
         required: true
         default: "yes"
       version:
-        description: "Override the release version number (e.g. 8.0.0a5)"
+        description: "Override the release version number (e.g. 5.0.0a5)"
 
 jobs:
   deploy-pypi:
     name: PyPI deployment
     runs-on: "ubuntu-latest"
-    if: github.event_name != 'push' || github.repository == 'DIRACGrid/DIRAC'
+    if: github.event_name != 'push' || github.repository == 'DIRACGrid/WebAppDIRAC'
     defaults:
       run:
         shell: bash -l {0}
@@ -28,6 +28,7 @@ jobs:
           git fetch --prune --unshallow
           git config --global user.email "ci@diracgrid.org"
           git config --global user.name "DIRACGrid CI"
+          git clone https://github.com/DIRACGrid/DIRAC.git ../DIRAC
       - uses: actions/setup-python@v2
         with:
           python-version: "3.9"
@@ -51,7 +52,7 @@ jobs:
           PREV_VERSION=$(git describe --tags --abbrev=0 --match '*[0-9].[0-9]*' --exclude 'v[0-9]r*' --exclude 'v[0-9][0-9]r*')
           REFERENCE_BRANCH=${GITHUB_REF#refs/heads/}
           echo "Making release notes for ${REFERENCE_BRANCH} since ${PREV_VERSION}"
-          ./docs/diracdoctools/scripts/dirac-docs-get-release-notes.py \
+          ../DIRAC/docs/diracdoctools/scripts/dirac-docs-get-release-notes.py \
               --token "${{ secrets.GITHUB_TOKEN }}" \
               --sinceTag "${PREV_VERSION}" \
               --branches "${REFERENCE_BRANCH}" \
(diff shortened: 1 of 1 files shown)


.github/workflows/deployment.yml
```diff
@@ -2,50 +2,117 @@ name: Deployment
...
+              > release.notes.new
+          cat release.notes.new
+      - name: Create tag if required
+        id: check-tag
+        run: |
+          set -xeuo pipefail
+          IFS=$'\n\t'
+          if [[ "${{ github.repository }}" == "DIRACGrid/WebAppDIRAC" ]]; then
+            if [[ "${{ github.event_name }}" == "workflow_dispatch" ]]; then
+              if [[ "${{ github.event.ref }}" =~ ^refs/heads/(integration|rel-v([8-9]|[1-9][0-9]+)\.[0-9]+)$ ]]; then
```
TaykYoku: here you mean integration, rel-v5.0, ...?

```suggestion
              if [[ "${{ github.event.ref }}" =~ ^refs/heads/(integration|rel-v([5-9]|[1-9][0-9]+)\.[0-9]+)$ ]]; then
```

.github/workflows/deployment.yml
```diff
@@ -2,50 +2,117 @@ name: Deployment
...
+              --repo "${{ github.repository }}" \
+              > release.notes.new
+          cat release.notes.new
+      - name: Create tag if required
+        id: check-tag
+        run: |
+          set -xeuo pipefail
+          IFS=$'\n\t'
+          if [[ "${{ github.repository }}" == "DIRACGrid/WebAppDIRAC" ]]; then
+            if [[ "${{ github.event_name }}" == "workflow_dispatch" ]]; then
```
TaykYoku, line 68: It seems I don't have enough rights to see how it work in `DIRACGrid / DIRAC`.
So now the standard way to create a release with GitHub UI will not work?
Do I understand correctly that to create a new release I need to go to "Actions" to select "deployment" workflow, then there should be a button "run workflow", where I specify branch and version, right?

.github/workflows/deployment.yml
```diff
@@ -2,50 +2,117 @@ name: Deployment
...
+                echo ::set-output name=create-release::true
+                echo ::set-output name=new-version::"$NEW_VERSION"
+              fi
+            fi
+          fi
       - name: Build distributions
-        run: python -m build
+        run: |
+          python -m build
+      - name: Make release on GitHub
```
TaykYoku, line 99: Maybe the [Marketplace](https://github.com/marketplace?category=&query=Release+sort%3Apopularity-desc&type=actions&verification=) already has a ready action for this purpose?

.github/workflows/deployment.yml
```diff
@@ -2,50 +2,117 @@ name: Deployment
...
     name: PyPI deployment
     runs-on: "ubuntu-latest"
     if: github.event_name != 'push' || github.repository == 'DIRACGrid/WebAppDIRAC'
+    defaults:
+      run:
+        shell: bash -l {0}
     steps:
       - uses: actions/checkout@v2
+        with:
+          token: ${{ secrets.PAT || github.token }}
```
TaykYoku, line 26: IIUC `PAT` more limited than standard `github.token`? I'm just trying to understand why just not use the standard one.

.github/workflows/deployment.yml
chrisburr: Yes, though there is no need to specify the version if it's a bump of the least significant digit (e.g. `5.0.0a5` will come after `5.0.0a4` but you have to manually specify it when `5.0.0` comes out).

Screenshot 2021-11-05 at 14 28 01

.github/workflows/deployment.yml
chrisburr: Possibly, when I looked for DIRACOS2 there wasn't anything suitable at the time. As I know this is working I'm inclined to leave this as-is.

.github/workflows/deployment.yml
chrisburr: `secrets.PAT` is less limited and able to push to protected braches

chrisburr: However it's not defined for PR builds so we fall back to the standard one when just testing the release building machinery.

.github/workflows/deployment.yml
TaykYoku: OK, if this works, then that's a good argument and I'm sure you've discussed a lot of different ways to do it.

Just for information, if you haven't paid attention to these products:
 - [GH Release](https://github.com/marketplace/actions/gh-release) just creating GitHub Releases
 - [Release Drafter](https://github.com/marketplace/actions/release-drafter) take care of release notes

TaykYoku:
Assuming that `DIRAC/docs/diracdoctools/scripts/dirac-docs-get-release-notes.py` and` DIRAC/.github/workflows/make_release.py` work as expected, the code looks good.
Probably the only way to test this is to merge and try.

chrisburr:
>  Probably the only way to test this is to merge and try.

Yeah unfortunately. I find testing deployment stuff is normally more work than just trying it (and even if you do test it there is normally something special about the real deployment environment and it fails regardless... 😆)

TaykYoku:
I will try to simulate this on my fork as much as possible and if I do not find a wrong behavior, then I will merge this PR.

PR #605, opened 2021-11-22
[5.0] remove Python 2 code

TaykYoku:
BEGINRELEASENOTES

CHANGE: remove Python 2 code

ENDRELEASENOTES


requirements.txt
```diff
@@ -3,7 +3,6 @@ diraccfg
 autopep8==1.3.3
 coverage
 requests>=2.9.1
-six
```
fstagni, line 6: is this file needed at all? I think we should align to use environment.yml like we do for DIRAC? Or none of the 2?

src/WebAppDIRAC/Core/HandlerMgr.py
```diff
@@ -18,16 +10,13 @@
...
-
 from WebAppDIRAC.Lib import Conf
 from WebAppDIRAC.Lib.WebHandler import WebHandler, _WebHandler, WebSocketHandler
 from WebAppDIRAC.Core.CoreHandler import CoreHandler
 from WebAppDIRAC.Core.StaticHandler import StaticHandler
 
 
-@six.add_metaclass(DIRACSingleton)
-class HandlerMgr(object):
+class HandlerMgr:
```
fstagni: You lost the metaclass. Why?

requirements.txt
TaykYoku: removed

src/WebAppDIRAC/Core/HandlerMgr.py
TaykYoku: well noticed, forgot to type it in python 3 style

fstagni (approved):

Issue #609, opened 2021-12-08
Accounting: insert text instead of using a pull-down menu for the JobGroup selector


fstagni:
(originally posted by @phicharp in ML)

Has anyone ever managed to make an accounting plot about jobs using the JobGroup selector? For me it blocks my portal page for a verrrrryyyyyyy long time and in the end nothing appears in the selector. This is probably due to the (too) large number of JobGroup values (i.e. probably all productions ever) but this is rather annoying:
1. because we may want to make such plots selecting specific productions
2. it blocks completely the application (and consumes 100% of a CPU)
JobGroup selector for job monitoring works fine but probably this is because here we only have the transformations that exist.

I would suggest to suppress the pulldown menu for this selector and allow the user to enter just a transformation number (not sure if this is possible) but as it is, we cannot use it…

TaykYoku:
I will note that the [Ext.form.field.Tag](https://docs.sencha.com/extjs/7.3.1/classic/Ext.form.field.Tag.html) element is involved here, it seems that it does not contain a ready-made solution, just as and for #425.

I'll see if it possible to implement something like regex.

> enter just a transformation number

Did not quite understand what kind of number, @fstagni if not difficult to add a link to it in the documentation or in the code.

fstagni:
> > enter just a transformation number
> 
> Did not quite understand what kind of number, @fstagni if not difficult to add a link to it in the documentation or in the code.

@phicharp is simply suggesting to suppress the pull-down menu, and use instead a free-text field.

TaykYoku:
@fstagni @phicharp I propose to consider the following solution #621

PR #621, opened 2022-01-31
[integration] resolve problem with more then 10000 items in a selector list Accounting

TaykYoku:
Issue #609

With this change if selector has more then 10000 items it comes as text field that allow you pass list of items:
```
dteam, admin, alytov
```

if you want to specify many values with the same prefix just set `*` at the end:
```
dteam, a*
```

BEGINRELEASENOTES

NEW: create textfield for selector if items more then 10000
NEW: allow to use `*` charter to set group values with the same prefix

ENDRELEASENOTES

### DIFF
--- a/src/WebAppDIRAC/WebApp/handler/AccountingHandler.py
+++ b/src/WebAppDIRAC/WebApp/handler/AccountingHandler.py
@@ -112,10 +112,11 @@ class AccountingHandler(WebHandler):
         callback["plotsList"] = data
         return {"success": "true", "result": callback}
 
-    def __parseFormParams(self, timeSelector, **pD):
+    def __parseFormParams(self, timeSelector, typeName=None, **pD):
         """Prepare parameters
 
         :param int timeSelector: time selector
+        :param str typeName: type of accounting data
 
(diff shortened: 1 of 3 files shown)


fstagni:
10000 is "just a large number", or did you go through some tests? Naively I think that even 1000 would be enough to trigger some slowness, but it's just a random guess.

TaykYoku:
At 10,000 it's somewhere around 2 seconds of waiting as I tested.

There is one point, I find it difficult to assess how convenient it is for the user to use the "list of values" / "simple field".

So what I did:
- I added the ability to arbitrarily enter values ​​and use * at the end as a designation of a group of values ​​(for ALL selectors), so theoretically this should speed up the selection of the required values, if they have the same prefix.
- if the values ​​are greater than N (now 10,000), the disadvantages become more severe than the advantages and the "list of values" becomes a "simple field".

I've considered a few more options, but they're a little trickier to implement, such as checking the boxes next to selectors to turn it into textfield, warning that the selector contains many values.

fstagni (approved):
I think it's fine.

PR #622, opened 2022-02-01
[v4r3] fix: encoding before hashing (py3)

fstagni:
BEGINRELEASENOTES

FIX: encoding before hashing (py3)

ENDRELEASENOTES


fstagni:
FYI: this fixes are necessary for when "save CSV data" of plots (both accounting and monitoring).

TaykYoku (approved):

PR #623, opened 2022-02-01
[v4r3] added .git-blame-ignore-revs file

fstagni:


.git-blame-ignore-revs
```diff
@@ -0,0 +1,3 @@
+# black commits
+ca412eba9b9c984ae04729de8f3dfc244b040e8e
+4ad573a81cd58f4d7917db60e05024626203a3c2
```
TaykYoku: No newline at end of file

```suggestion
4ad573a81cd58f4d7917db60e05024626203a3c2

```

TaykYoku (approved):

.git-blame-ignore-revs
```diff
@@ -0,0 +1,3 @@
+# black commits
```
TaykYoku, line 1: Perhaps it is worth paraphrasing, because it can be taken ambiguously
```suggestion
# Adoption of coding style used by Black
```

PR #610, opened 2022-01-06
[integration] update WebHandler (depended #5399)

TaykYoku:
BEGINRELEASENOTES
Updates and optimisations due to proposed changes to BaseRequestHandler, see https://github.com/DIRACGrid/DIRAC/pull/5399

Changes tested on lbcertifdiracoauth.cern.ch.

CHANGE: Updates and optimisations due to proposed changes to BaseRequestHandler

ENDRELEASENOTES

PR #632, opened 2022-02-10
[5.0] continue #610, fix bugs, docs

TaykYoku:
`SiteSummary` and `ResourceSummary` handlers use common methods to avoid duplication and unexpected behavior of one of them depending on the other `SummaryHandlerMix` was created, which describes common methods.

**The bug** is that the result of helper methods should be returned using `return`, not `self.finish`. It is also not necessary to use the combination `yield self.threadTask` in the `SummaryHandlerMix` as all auxiliary methods are already called in the non-main thread, see `web_action`.

BEGINRELEASENOTES

*ResourceSummary/SiteSummary
FIX: docs, optimize, fix bugs

ENDRELEASENOTES

### DIFF
--- a/src/WebAppDIRAC/WebApp/handler/ResourceSummaryHandler.py
+++ b/src/WebAppDIRAC/WebApp/handler/ResourceSummaryHandler.py
@@ -13,7 +13,7 @@ class SummaryHandlerMix(WebHandler):
     AUTH_PROPS = "all"
     ELEMENT_TYPE = None
 
-    def _getSelectionData(self):
+    def _getSelectionData(self) -> dict:
         """It returns the possible selection data"""
         callback = {"name": set(), "elementType": set(), "status": set(), "statusType": set(), "tokenOwner": set()}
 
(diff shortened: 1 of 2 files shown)


src/WebAppDIRAC/WebApp/handler/ResourceSummaryHandler.py
```diff
@@ -148,24 +157,18 @@ def setToken(self, requestParams):
...
-
-        pub = PublisherClient()
+        :param requestParams: request parameters
+        """
+        if error := self.__checkAuth(username := self.getUserName()):
+            return {"success": "false", "error": error}
 
-        res = yield self.threadTask(
-            pub.setStatus,
+        res = PublisherClient().setStatus(
```
fstagni, line 171: What was wrong with the previous implementation?

TaykYoku: This place in the code already in a second (non-main) thread.

```python
class ResourceSummaryHandler(SummaryHandlerMix):
    ...
    @asyncGen
    def web_action(self):
        callback = yield self.threadTask(self._action)  # this line run all sync logic in a second thread
        self.finish(callback)
```
`web_action` run in thread `_action` that run `setStatus` that run `PublisherClient().setStatus(..`

TaykYoku:
can I merge this?

fstagni (approved):

PR #633, opened 2022-02-10
[5.0] pass file name as str to FileResponse

TaykYoku:
BEGINRELEASENOTES

*Accounting
FIX: pass file name as str to FileResponse

ENDRELEASENOTES

### DIFF
--- a/src/WebAppDIRAC/Lib/WebHandler.py
+++ b/src/WebAppDIRAC/Lib/WebHandler.py
@@ -75,18 +75,20 @@ class FileResponse(TornadoResponse):
         return FileResponse(data, 'filename', 'csv')
     """
 
-    def __init__(self, payload=None, fileName=None, ext=None, cache=True):
+    def __init__(self, payload, fileName: str, ext: str = "", cache: bool = True):
         """C'or
 
         :param payload: response body
-        :param str fileName: CSV name
-        :param str ext: file type
-        :param bool cache: use cache
+        :param fileName: CSV name
+        :param ext: file type
+        :param cache: use cache
         """
-        self.name, self.ext = os.path.splitext(fileName)
-        self.ext = (ext or self.ext).lower()
+        name, _ext = os.path.splitext(fileName)
+        self.ext = (ext or _ext).lower()
+        # Generate file name
+        self.fileHash = md5(name.encode()).hexdigest()  # MD5 take a bytes
         self.cache = cache
-        super(FileResponse, self).__init__(payload, 200)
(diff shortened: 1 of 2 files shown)


src/WebAppDIRAC/WebApp/handler/AccountingHandler.py
```diff
@@ -229,7 +229,7 @@ def web_getPlotImg(self, file=None, **kwargs):
 
         tempFile.seek(0)
         data = tempFile.read()
-        return FileResponse(data, plotImageFile.encode(), "png")
+        return FileResponse(data, plotImageFile, "png")
```
chrisburr, line 232: Can you explain this change? I'm suspicious that you're seeing a difference between how DEncode/JEncode behave.

TaykYoku: This change does not apply to `DEncode`/ `JEncode`.
The file name is obtained as `str`, and `FileResponse` also expects the type `str`, so `.encode()` is not required here.

`FileResponse` will convert this to bytes as needed(e.g.: for **md5**)

TaykYoku: Additional explanations may be needed here. WebApp handlers inherit `TornadoREST` and when you describe an argument in the target method (eg: `def web_cats (self, color):`) then you will have access to this value. The value type corresponds to the annotation (by default its `str`).

chrisburr: Ah right, yes I misunderstood the code 👍

src/WebAppDIRAC/Lib/WebHandler.py
```diff
@@ -75,18 +75,20 @@ def web_myMethod(self):
         return FileResponse(data, 'filename', 'csv')
     """
 
-    def __init__(self, payload=None, fileName=None, ext=None, cache=True):
+    def __init__(self, payload=None, fileName: str = None, ext: str = None, cache: bool = True):
```
chrisburr: This is the wrong type hint, it should be:
```suggestion
    def __init__(self, payload=None, fileName: Optional[str] = None, ext: Optional[str] = None, cache: bool = True):
```
Where `Optional` comes from `from typing import Optional`. In Python 3.10+ it can be `None | str`.

TaykYoku: Good point, fixed

fstagni (approved):

PR #641, opened 2022-03-21
Removing ComponentMonitoring from WebApp

rupozzi:
Removing Component Monitoring from web app following [#5944](https://github.com/DIRACGrid/DIRAC/pull/5944).

BEGINRELEASENOTES

CHANGE: removed Component Monitoring

ENDRELEASENOTES

### DIFF
--- a/src/WebAppDIRAC/WebApp/static/DIRAC/Accounting/classes/Accounting.js
+++ b/src/WebAppDIRAC/WebApp/static/DIRAC/Accounting/classes/Accounting.js
@@ -91,16 +91,6 @@ Ext.define("DIRAC.Accounting.classes.Accounting", {
             ["JobSplitType", "Job Split Type"],
           ],
         },
-        ComponentMonitoring: {
-          title: "Component Monitoring",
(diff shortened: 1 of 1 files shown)


src/WebAppDIRAC/WebApp/static/DIRAC/Accounting/classes/Accounting.js
```diff
@@ -91,16 +91,6 @@ Ext.define("DIRAC.Accounting.classes.Accounting", {
             ["JobSplitType", "Job Split Type"],
           ],
         },
-        ComponentMonitoring: {
```
fstagni, line 94: This should stay, I think.

rupozzi: We won't have a ComponentMonitoring anymore after [#5944](https://github.com/DIRACGrid/DIRAC/pull/5944) gets merged, so unless we replace it with a service- and agentmonitoring in the WebApp I think it needs to be removed.

PR #642, opened 2022-03-22
Remove activity monitor from web app

rupozzi:
Removing activity monitor following Pull Request [#5969](https://github.com/DIRACGrid/DIRAC/pull/5965).

BEGINRELEASENOTES

CHANGE: removed activity monitor 

ENDRELEASENOTES


src/WebAppDIRAC/WebApp/handler/MonitoringHandler.py
```diff
@@ -1,257 +0,0 @@
-import os
```
fstagni: This should stay!

chrisburr: It does need modifying though to remove things like: `MonitoringClient().getActivitiesContents`

rupozzi: @chrisburr I don't find that in MonitoringHandler... I think it's activityMonitoringHandler

rupozzi:
Is activityMonitoringHandler.py to be deleted?

fstagni:
> Is activityMonitoringHandler.py to be deleted?

yes, but not with this PR.

PR #649, opened 2022-05-13
[5.0] use ComponentMonitoringClient instead of ComponentMonitoringDB

fstagni:
BEGINRELEASENOTES

FIX: use ComponentMonitoringClient instead of deprecated ComponentMonitoringDB in SystemAdministrationHandler

ENDRELEASENOTES

### DIFF
--- a/src/WebAppDIRAC/WebApp/handler/SystemAdministrationHandler.py
+++ b/src/WebAppDIRAC/WebApp/handler/SystemAdministrationHandler.py
@@ -2,14 +2,12 @@
 """
 
 import json
-import datetime
 
 from DIRAC import gConfig, gLogger
 from DIRAC.Core.Utilities.List import uniqueElements
 from DIRAC.FrameworkSystem.Client.NotificationClient import NotificationClient
 from DIRAC.FrameworkSystem.Client.SystemAdministratorClient import SystemAdministratorClient
 from DIRAC.FrameworkSystem.Client.ComponentMonitoringClient import ComponentMonitoringClient
-from DIRAC.FrameworkSystem.DB.ComponentMonitoringDB import ComponentMonitoringDB
 
 from WebAppDIRAC.Lib.WebHandler import WebHandler, asyncGen
 
@@ -42,6 +40,7 @@ class SystemAdministrationHandler(WebHandler):
             )
 
         self.finish({"success": "true", "result": sorted(callback, key=lambda i: i["Host"]), "total": len(callback)})
+        return
 
     @asyncGen
     def web_getHostData(self):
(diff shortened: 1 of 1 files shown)


fstagni:
Review, please

src/WebAppDIRAC/WebApp/handler/SystemAdministrationHandler.py
```diff
@@ -636,101 +633,30 @@ def web_getSelectionData(self):
...
-            componentNames = list(json.loads(self.get_argument("ComponentName")))
-
         componentModules = []
         if "ComponentModule" in self.request.arguments:
             componentModules = list(json.loads(self.get_argument("ComponentModule")))
 
-        showAll = 0
-        if "showAll" in self.request.arguments:
-            showAll = int(self.get_argument("showAll"))
-
```
TaykYoku, line 661: Do web portal no longer need option "showAll"?
The showAll option is also described in /WebAppDIRAC/WebApp/static/DIRAC/SystemAdministration/classes/SystemAdministration.js

src/WebAppDIRAC/WebApp/handler/SystemAdministrationHandler.py
```diff
@@ -636,101 +633,30 @@ def web_getSelectionData(self):
...
-                            if module != "" and module in componentModules:
-                                fullNames += [path]
-                            elif module == "" and j in componentModules:
-                                fullNames += [path]
 
-                            compMatching[path] = module if module != "" else path
+        condDict = {"Type": componentTypes, "DIRACModule": componentModules}
+        gLogger.debug("condDict" + str(condDict))
 
+        res = ComponentMonitoringClient.getComponents(condDict, True, True)
```
TaykYoku:
```suggestion
        res = ComponentMonitoringClient().getComponents(condDict, True, True)
```

src/WebAppDIRAC/WebApp/handler/SystemAdministrationHandler.py
```diff
@@ -636,101 +633,30 @@ def web_getSelectionData(self):
...
+        if not res["OK"]:
+            self.finish({"success": "false", "error": res["Message"]})
```
TaykYoku, line 653: If you do not return at this point, the code will continue to run until the end of the function and will fail with `KeyError: 'Value'`

```suggestion
            self.finish({"success": "false", "error": res["Message"]})
            return
```

src/WebAppDIRAC/WebApp/handler/SystemAdministrationHandler.py
fstagni: I removed it because `showAll` is unused in this module.

PR #656, opened 2022-06-14
[5.0] remove UserCredentials, minor fixes

TaykYoku:
BEGINRELEASENOTES

FIX: Bug fixes
CHANGE: remove unused UserCredentials

ENDRELEASENOTES

Issue #658, opened 2022-06-17
Replace WebHandler with _WebHandler


chrisburr:
Apologies if I haven't found where this is was discussed but it seems that we currently have two implementations of the `WebHandler`:

https://github.com/DIRACGrid/WebAppDIRAC/blob/d70c268ff47fc051ca29fd0415fcaeb0e7587f2a/src/WebAppDIRAC/Lib/WebHandler.py#L161

https://github.com/DIRACGrid/WebAppDIRAC/blob/d70c268ff47fc051ca29fd0415fcaeb0e7587f2a/src/WebAppDIRAC/Lib/WebHandler.py#L401

Most classes seem to be using `_WebHandler` which appears to be a much nicer implementation which delegates every request to a dedicated thread so the `@asyncGen` and ` yield self.threadTask` stuff isn't needed like in:

https://github.com/DIRACGrid/WebAppDIRAC/blob/d70c268ff47fc051ca29fd0415fcaeb0e7587f2a/src/WebAppDIRAC/WebApp/handler/SystemAdministrationHandler.py#L91-L92

https://github.com/DIRACGrid/WebAppDIRAC/blob/d70c268ff47fc051ca29fd0415fcaeb0e7587f2a/src/WebAppDIRAC/WebApp/handler/SystemAdministrationHandler.py#L104

Is there a reason why we need both implementations? I think we should just move the legacy style handlers to the new one and delete the old implementation.

TaykYoku:
Yes, you are absolutely right, now the process of switching to `_WebHandler` is underway, as soon as it ends `WebHandler` is simply removed and `_WebHandler` will be renamed to `WebHandler`.

chrisburr:
I think it would be good to deal with this before releasing v5.0 so we have a cleaner starting point fot v5.1 (which I have some ideas for).

Quickly checking these are the modules which need to be updated in vanilla DIRAC. I'll try to start working through them now starting from the top and checking them off:

- [x] `src/WebAppDIRAC/WebApp/handler/ConfigurationManagerHandler.py`
- [x] `src/WebAppDIRAC/WebApp/handler/ProxyManagerHandler.py`
- [x] `src/WebAppDIRAC/WebApp/handler/ProxyUploadHandler.py`
- [x] `src/WebAppDIRAC/WebApp/handler/PublicStateManagerHandler.py`
- [x] `src/WebAppDIRAC/WebApp/handler/RegistryManagerHandler.py`
[...]

PR #659, opened 2022-06-17
[5.0] Set ThreadConfig correctly in WebHandler and _WebHandler

chrisburr:
Requires https://github.com/DIRACGrid/DIRAC/pull/6174.
Supersedes https://github.com/DIRACGrid/WebAppDIRAC/pull/656 and https://github.com/DIRACGrid/DIRAC/pull/6142.

BEGINRELEASENOTES

FIX: Sending extraCredentials when using X509 based RPC calls on behalf of users.
CHANGE: The X-RefreshConfiguration header can no longer be used to force a refresh of the CS.

ENDRELEASENOTES

### DIFF
--- a/src/WebAppDIRAC/Lib/WebHandler.py
+++ b/src/WebAppDIRAC/Lib/WebHandler.py
@@ -1,6 +1,3 @@
-""" Main module
-"""
-
 import os
 import re
 import json
@@ -36,32 +33,6 @@ from WebAppDIRAC.Lib.SessionData import SessionData
 global gThreadPool
 gThreadPool = ThreadPoolExecutor(100)
 sLog = gLogger.getSubLogger(__name__)
-
-
-class UserCredentials:
-    """With this class you can make RPC calls using user credantials
-
-    Usage:
-
-        def post_job(self):
-            with UserCredentials(**self.credDict):
-                # Do stuff
-
-    """
-
-    __disetConfig = ThreadConfig()
-
-    def __init__(self, **kwargs):
-        self.dn = kwargs["DN"]
-        self.group = kwargs["group"]
-        self.setup = kwargs["setup"]
-
-    def __enter__(self):
-        self.__disetConfig.load((self.dn, self.group, self.setup))
-        return self.__disetConfig
-
-    def __exit__(self, type, value, traceback):
-        self.__disetConfig.reset()
 
 
 class FileResponse(TornadoResponse):
@@ -232,7 +203,6 @@ class _WebHandler(TornadoREST):
         We make the assumption that there is always going to be a ``method`` argument
         regardless of the HTTP method used
         """
-
         # Parse request URI
         match = self.PATH_RE.match(self.request.path)
         groups = match.groups()
@@ -277,13 +247,12 @@ class _WebHandler(TornadoREST):
 
     def _gatherPeerCredentials(self):
         """
-        Load client certchain in DIRAC and extract informations.
+        Load client certificate chain in DIRAC and extract informations.
 
         The dictionary returned is designed to work with the AuthManager,
         already written for DISET and re-used for HTTPS.
 
-        :returns: a dict containing the return of :py:meth:`DIRAC.Core.Security.X509Chain.X509Chain.getCredentials`
-                  (not a DIRAC structure !)
+        :returns: dict containing the return of :py:meth:`DIRAC.Core.Security.X509Chain.X509Chain.getCredentials`
         """
         # Authorization type
         self.__authGrant = ["VISITOR"]
@@ -301,7 +270,7 @@ class _WebHandler(TornadoREST):
         return credDict
 
     def _authzSESSION(self):
-        """Fill credentionals from session
+        """Fill credentials from session
 
         :return: S_OK(dict)
         """
@@ -433,7 +402,7 @@ class WebHandler(_WebHandler):
         return IOLoop.current().run_in_executor(gThreadPool, functools.partial(threadJob, *targs, **kwargs))
 
     def finish(self, data=None, *args, **kwargs):
-        """Finishes this response, ending the HTTP request. More detailes:
+        """Finishes this response, ending the HTTP request. More details:
         https://www.tornadoweb.org/en/stable/_modules/tornado/web.html#RequestHandler.finish
         """
         if data and isinstance(data, dict):
--- a/src/WebAppDIRAC/Lib/WebHandler.py
+++ b/src/WebAppDIRAC/Lib/WebHandler.py
@@ -1,3 +1,4 @@
+import contextlib
 import os
 import re
 import json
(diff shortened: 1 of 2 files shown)


chrisburr:
I've tested this for both `lbcertifdirac70` and `lhcb-cert-dirac` and everything seems fine.

chaen (approved):

src/WebAppDIRAC/Lib/WebHandler.py
```diff
@@ -422,18 +367,14 @@ def post(self, *args, **kwargs):
...
+            with self._setupThreadConfig():
+                return method(*args, **kwargs)
 
-        targs = (args, self.__disetDump)
-        return IOLoop.current().run_in_executor(gThreadPool, functools.partial(threadJob, *targs, **kwargs))
+        return IOLoop.current().run_in_executor(gThreadPool, functools.partial(threadJob, *args, **kwargs))
 
     def finish(self, data=None, *args, **kwargs):
-        """Finishes this response, ending the HTTP request. More detailes:
+        """Finishes this response, ending the HTTP request. More detiles:
```
fstagni:
```suggestion
        """Finishes this response, ending the HTTP request. More details:
```

src/WebAppDIRAC/Lib/WebHandler.py
```diff
@@ -422,18 +367,14 @@ def post(self, *args, **kwargs):
...
     def threadTask(self, method, *args, **kwargs):
-        def threadJob(*targs, **tkwargs):
-            args = targs[0]
-            disetConf = targs[1]
-            self.__disetConfig.reset()
-            self.__disetConfig.load(disetConf)
-            return method(*args, **tkwargs)
+        def threadJob(*args, **tkwargs):
+            with self._setupThreadConfig():
+                return method(*args, **kwargs)
```
fstagni, line 372: while `method` is not a reserved keyword, maybe it's a bit too generic...?

src/WebAppDIRAC/Lib/WebHandler.py
```diff
@@ -422,18 +367,14 @@ def post(self, *args, **kwargs):
...
 
-        targs = (args, self.__disetDump)
```
fstagni, line 432: why is this `self.__disetDump` not useful (anymore)?

src/WebAppDIRAC/Lib/WebHandler.py
```diff
@@ -345,14 +293,11 @@ def _authzSESSION(self):
...
 
-    def getCurrentSession(self):
-        return self.__session
-
     def getUserSetup(self):
         return self.__setup
 
     def getSessionData(self):
-        if not self.__sessionData:
+        if not hasattr(self, "__sessionData"):
```
fstagni, line 300: is this one of those points where `ObjectLoader` should be used?

src/WebAppDIRAC/Lib/WebHandler.py
chrisburr: This entire `WebHandler` class should be removed entirely: https://github.com/DIRACGrid/WebAppDIRAC/issues/658

src/WebAppDIRAC/Lib/WebHandler.py
chrisburr: The current user can be queried from `self` directly so I see no reason to duplicate this information into `self.__disetDump`. It's also a thing that only applies to the legacy implementation which should be removed: https://github.com/DIRACGrid/WebAppDIRAC/issues/658

src/WebAppDIRAC/Lib/WebHandler.py
chrisburr: I don't think so?

src/WebAppDIRAC/Lib/WebHandler.py
```diff
@@ -422,18 +367,14 @@ def post(self, *args, **kwargs):
         return self.get(*args, **kwargs)
 
     def threadTask(self, method, *args, **kwargs):
-        def threadJob(*targs, **tkwargs):
-            args = targs[0]
-            disetConf = targs[1]
-            self.__disetConfig.reset()
-            self.__disetConfig.load(disetConf)
-            return method(*args, **tkwargs)
+        def threadJob(*args, **tkwargs):
```
TaykYoku:
```suggestion
        def threadJob(*args, **kwargs):
```

src/WebAppDIRAC/Lib/WebHandler.py
```diff
@@ -17,16 +15,11 @@
...
 
-from DIRAC import gLogger, gConfig, S_OK, S_ERROR
+from DIRAC import gLogger, S_OK, S_ERROR
 from DIRAC.Core.Utilities.JEncode import DATETIME_DEFAULT_FORMAT
 from DIRAC.Core.Utilities.Decorators import deprecated
-from DIRAC.Core.DISET.AuthManager import AuthManager
 from DIRAC.Core.DISET.ThreadConfig import ThreadConfig
-from DIRAC.Core.Tornado.Server.TornadoREST import (
-    TornadoREST,
-    authorization,
```
TaykYoku, line 27: Just for note: I added this import for the convenience of writing a handler to get everything you need from one place 
```python
from WebAppDIRAC.Lib.WebHandler import ...
```

chrisburr: It was currently only used in one place?

Though in gernerla I agree with the idea of having more intentional imports but, if we do this, we should start setting `__all__` so it's clear what is supposed to be part of the interface.

src/WebAppDIRAC/Lib/WebHandler.py
```diff
@@ -232,58 +195,43 @@ async def prepare(self):
...
+            threadConfig.setDN(userDN)
+        if userGroup := self.getUserGroup():
+            threadConfig.setGroup(userGroup)
+        threadConfig.setSetup(self.__setup)
+        try:
+            yield
+        finally:
+            threadConfig.reset()
 
-    def __forceRefreshCS(self):
```
TaykYoku, line 268: I think it is worth adding this deletion to the release notes

chrisburr: THis has never been in a (non-alpha) release has it and I couldn't find any documentation for it?

TaykYoku: There is no documentation, this is a specific function for force refresh CS, when adding it an entry was left in release.notes, so deleting it should also be reflected in release.notes (something like "__forceRefreshCS was removed").

---
I do not see the need to leave it as it is, so I agree with its removal. If the force refresh CS request mechanism is needed, I think it should be used at the `BaseRequestHandler` level as the core class.

chrisburr: Ah, I see it was added in v4r1p4 but the divergence in the git history made it appear to be something new.

I've updated the release notes to reflect that `X-RefreshConfiguration` is no longer respected as that's the only observable behavour (hopefully nobody is calling ~~`__WebHandler__forceRefreshCS`, which actually might not even be possible to call as it will be resolved to the non-existant `_CLASSNAME__WebHandler__forceRefreshCS`~~ actually it's `_WebHandler__forceRefreshCS` as [leading underscores](https://docs.python.org/3/tutorial/classes.html#private-variables) are stripped).

TaykYoku: This is an internal method, it is part of the algorithm that runs during each request, look tornadoes `prepare()` method, Its also described in the `BaseRequestHandler`.

It depended https://github.com/DIRACGrid/DIRAC/pull/4241. In fact, I would not touch it yet and would consider it later, as a less priority.

TaykYoku: @chrisburr Could you please not remove `__forceRefreshCS` method in this PR, as it is a continuation of the mechanism developed in https://github.com/DIRACGrid/DIRAC/pull/4241. If you think it is necessary to delete it, create a separate PR, so as not to delay the acceptance of this one.

chrisburr: It's no longer used in DIRAC 8.0 is it? At least I couldn't find it?

TaykYoku: so you're right, this logic is gone.. well OK, then I merge this PR

PR #638, opened 2022-02-27
[5.0] add BaseRequestHandler to MonitoringHandler

TaykYoku:
BEGINRELEASENOTES

NEW: add BaseRequestHandler to MonitoringHandler

ENDRELEASENOTES

PR #660, opened 2022-06-17
[5.0] Modernise asyncGen and only use new _WebHandler class

### LINKED ISSUE #658: Replace WebHandler with _WebHandler

chrisburr:
Fixes https://github.com/DIRACGrid/WebAppDIRAC/issues/658

Draft as it is untested and needs #659 

BEGINRELEASENOTES

CHANGE: Modernise asyncGen and only use new _WebHandler class

ENDRELEASENOTES

### DIFF
--- a/src/WebAppDIRAC/Core/HandlerMgr.py
+++ b/src/WebAppDIRAC/Core/HandlerMgr.py
@@ -10,7 +10,7 @@ from DIRAC.Core.Utilities.DIRACSingleton import DIRACSingleton
 from DIRAC.Core.Utilities.Extensions import extensionsByPriority, getExtensionMetadata
 
 from WebAppDIRAC.Lib import Conf
-from WebAppDIRAC.Lib.WebHandler import WebHandler, _WebHandler
+from WebAppDIRAC.Lib.WebHandler import _WebHandler
 from WebAppDIRAC.Core.StaticHandler import StaticHandler
 
 
@@ -61,11 +61,10 @@ class HandlerMgr(metaclass=DIRACSingleton):
         ol = ObjectLoader()
         handlerList = []
         self.log.debug("Add handles from: %s", self.__handlersLocation)
-        for parentClass in [WebHandler, _WebHandler]:
-            result = ol.getObjects(self.__handlersLocation, parentClass=parentClass, recurse=True, continueOnError=True)
-            if not result["OK"]:
-                return result
-            handlerList += list(result["Value"].items())
+        result = ol.getObjects(self.__handlersLocation, parentClass=_WebHandler, recurse=True, continueOnError=True)
+        if not result["OK"]:
+            return result
+        handlerList += list(result["Value"].items())
         self.__handlers = collections.OrderedDict(handlerList)
 
         # ['/opt/dirac/pro/WebAppExt/WebApp/static', ...]
--- a/src/WebAppDIRAC/Lib/WebHandler.py
+++ b/src/WebAppDIRAC/Lib/WebHandler.py
@@ -1,6 +1,3 @@
-""" Main module
-"""
-
 import os
 import re
 import json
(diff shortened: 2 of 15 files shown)


TaykYoku:
I would suggest splitting this PR, one PR per handler to speed up viewing and acceptance.

chrisburr:
At the moment it's broken but I'll try to fix it on Monday.

> I would suggest splitting this PR, one PR per handler to speed up viewing and acceptance.

I don't think this would actually help as excluding `WebHandler.py` and `HandlerMgr.py` the changes are very mechanical and just removing `@asyncGen`+`yield self.threadTask` and changing `_WebHandler` to `WebHandler`.

That said, I can make the removal of `asyncGen` and `WebHandler` a separate PR once it's working.

TaykYoku:
This is not entirely true, there are some nuances. The very idea of creating `_WebHandler` based on `BaseRequestHandler` is to change the style of writing handlers. In addition to removing `@asyncGen` + `yield self.threadTask` and changing `_WebHandler` to `WebHandler`, for ex. you need to use `return` instead of `self.finish`, describe **request arguments** via **function annotations**.

I would suggest solving https://github.com/DIRACGrid/WebAppDIRAC/pull/659 first.

Using your list https://github.com/DIRACGrid/WebAppDIRAC/issues/658#issuecomment-1158915346, create a PR by PR for each hadler (as was done before, eg #639, #638 ...). IMHO it's so much easier to test and review, and it can be done in parallel, I can go from the bottom of the list.

`WebHandler` is removed last when all handlers have been ported to `_WebHandler` and tested.

What do you think?

chrisburr:
> This is not entirely true, there are some nuances. The very idea of creating `_WebHandler` based on `BaseRequestHandler` is to change the style of writing handlers. In addition to removing `@asyncGen` + `yield self.threadTask` and changing `_WebHandler` to `WebHandler`, for ex. you need to use `return` instead of `self.finish`, describe **request arguments** via **function annotations**.

Where are the required changes documented? It's really important as it also needs to be done in extensions.

> I would suggest solving #659 first.

I think this is already solved modulo review.

> Using your list [#658 (comment)](https://github.com/DIRACGrid/WebAppDIRAC/issues/658#issuecomment-1158915346), create a PR by PR for each hadler (as was done before, eg #639, #638 ...). IMHO it's so much easier to test and review, and it can be done in parallel, I can go from the bottom of the list.
>
> `WebHandler` is removed last when all handlers have been ported to `_WebHandler` and tested.

As mentioned in https://github.com/DIRACGrid/WebAppDIRAC/issues/658 I think this needs to be done before releasing 5.0 and DIRAC 8.0, which implies it needs to be done before Thursday as that's supposed to be the final hackathon. Once we're done I think we should also rename `_WebHandler` to something else (possibly `WebHandler`) as it currenly looks like an implementation detail.

PR #662, opened 2022-06-17
[5.0] TransfomationMonitor use BaseRequestHandler

TaykYoku:
issue https://github.com/DIRACGrid/WebAppDIRAC/issues/658

BEGINRELEASENOTES

CHANGE: TransformationMonitor use BaseRequestHandler

ENDRELEASENOTES

### DIFF
--- a/src/WebAppDIRAC/WebApp/handler/TransformationMonitorHandler.py
+++ b/src/WebAppDIRAC/WebApp/handler/TransformationMonitorHandler.py
@@ -327,27 +327,29 @@ class TransformationMonitorHandler(WebHandler):
             return {"success": "false", "error": result["Message"]}
 
         data = result["Value"]
-        if "TotalRecords" in data and (total := data["TotalRecords"]) > 0:
-            if "ParameterNames" in data and "Records" in data:
-                if (headLength := len(head := data["ParameterNames"])) > 0:
-                    if len(data["Records"]) > 0:
-                        callback = []
-                        for job in data["Records"]:
-                            callback.append({head[j]: job[j] for j in range(headLength)})
-                        timestamp = datetime.datetime.utcnow().strftime("%Y-%m-%d %H:%M [UTC]")
-                        if "Extras" in data:
-                            return {
-                                "success": "true",
-                                "result": callback,
-                                "total": total,
-                                "extra": data["Extras"],
-                                "date": timestamp,
-                            }
-                        return {"success": "true", "result": callback, "total": total, "date": timestamp}
-                    return {"success": "false", "result": "", "error": "There are no data to display"}
-                return {"success": "false", "result": "", "error": "ParameterNames field is undefined"}
+        total = len(data.get("TotalRecords", []))
+        if total == 0:
+            return {"success": "false", "result": "", "error": "There were no data matching your selection"}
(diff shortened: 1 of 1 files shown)


src/WebAppDIRAC/WebApp/handler/TransformationMonitorHandler.py
```diff
@@ -8,512 +8,417 @@
...
+        if not result["OK"]:
+            return json.dumps({"success": "false", "error": result["Message"]})
+
+        data = result["Value"]
+
+        if "TotalRecords" not in data:
+            return json.dumps({"success": "false", "result": "", "error": "Data structure is corrupted"})
+
+        if (total := data["TotalRecords"]) < 1:
+            return json.dumps({"success": "false", "result": "", "error": "There were no data matching your selection"})
```
chrisburr: Is the `json.dumps` needed? Or could the dictionary be returned directly?

src/WebAppDIRAC/WebApp/handler/TransformationMonitorHandler.py
```diff
@@ -8,512 +8,417 @@
...
+                                "result": callback,
+                                "total": total,
+                                "extra": data["Extras"],
+                                "date": timestamp,
+                            }
+                        return {"success": "true", "result": callback, "total": total, "date": timestamp}
+                    return {"success": "false", "result": "", "error": "There are no data to display"}
+                return {"success": "false", "result": "", "error": "ParameterNames field is undefined"}
+            return {"success": "false", "result": "", "error": "Data structure is corrupted"}
+        return {"success": "false", "result": "", "error": "There were no data matching your selection"}
```
chrisburr: I'm ignoring most of style comments but I think this function is bad enough that we should fix it:

```suggestion
        total = len(data.get("TotalRecords", []))
        if total == 0:
            return {"success": "false", "result": "", "error": "There were no data matching your selection"}
        if not ("ParameterNames" in data and "Records" in data):
            return {"success": "false", "result": "", "error": "Data structure is corrupted"}
        head = data.get("ParameterNames", [])
        if len(head) == 0:
            return {"success": "false", "result": "", "error": "ParameterNames field is undefined"}
        if len(data["Records"]) == 0:
            return {"success": "false", "result": "", "error": "There are no data to display"}
        callback = []
        for job in data["Records"]:
            callback.append(dict(zip(head, job)))
        timestamp = datetime.datetime.utcnow().strftime("%Y-%m-%d %H:%M [UTC]")
        if "Extras" in data:
            return {
                "success": "true",
                "result": callback,
                "total": total,
                "extra": data["Extras"],
                "date": timestamp,
            }
        return {"success": "true", "result": callback, "total": total, "date": timestamp}     
```

chrisburr: Also I think we should be a little more selective with the use of the `:=` operator, things like this are quite tricky to parse:

```
if (headLength := len(head := data["ParameterNames"])) > 0:
```

src/WebAppDIRAC/WebApp/handler/TransformationMonitorHandler.py
TaykYoku: No, `BaseRequestHandler` already has encoder. Fixed

PR #677, opened 2022-06-23
[5.0] set correct argument type for _expand

TaykYoku:
BEGINRELEASENOTES

FIX: set correct argument type for _expand

ENDRELEASENOTES


TaykYoku:
[Verify RSS: use RSS 'Active'](https://trello.com/c/PAYfENSL/14-verify-rss-use-rss-active)

fstagni (approved):

Issue #678, opened 2022-07-12
How to extend new-style webapp handlers?


chrisburr:
The new style webapp handlers that were added have made it so that parameters are now defined as arguments to the function:

https://github.com/DIRACGrid/WebAppDIRAC/blob/278c231107c8325f07fb332cb52f60765870c5b5/src/WebAppDIRAC/WebApp/handler/TransformationMonitorHandler.py#L79-L97

In LHCb we overload several of these methods to add additional parameters, for example on this method we would also have a `hot=False` parameter that should be included in the search dict. We previously did this by overloading the `_request` function with something like:

```python
    def _request(self):
        req = super()._request(*args, **kwargs)
        hotFlag = json.loads(self.get_argument("Hot", "[false]"))[-1]
        if hotFlag:
            req["Hot"] = hotFlag
        return req
```

However the new style has changed to have `__prepareParameters`. While I can easily re-name it to made it overloadable I don't see an nice way of being able to add parameters with completely copying the code from `WebAppDIRAC`.

PR #680, opened 2022-08-03
[5.0] Make TransformationMonitorHandler.__prepareParameters less private

chrisburr:
See https://github.com/DIRACGrid/WebAppDIRAC/issues/678 for the motivation. I think this solution is good enough to fix the one use case I'm currently aware of.

### DIFF
--- a/src/WebAppDIRAC/WebApp/handler/TransformationMonitorHandler.py
+++ b/src/WebAppDIRAC/WebApp/handler/TransformationMonitorHandler.py
@@ -100,7 +100,7 @@ class TransformationMonitorHandler(WebHandler):
         if self.getUserName().lower() == "anonymous":
             return {"success": "false", "error": "You are not authorised"}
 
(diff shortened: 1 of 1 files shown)


fstagni:
This would fix our (LHCb) known use case, is there anything more generic that can be done?

chrisburr:
> is there anything more generic that can be done?

Not that I can reasonably think of. I suspect the proper solution for even us is to remove the `hot` flag as I'm not sure what it even does? (though I'm not proposing we do that right now)

chrisburr:
As it stands the limitations for extensions are the same as in the standard service handlers. It's just a bit more messy for the web app as the methods have a lot of URL parameteres.

Longer term I have some ideas for vanilla DIRAC to improve the `types_METHOD` propery that will remove the issue but that's not for DIRAC v8.0

fstagni (approved):

PR #690, opened 2022-10-05
[5.0] ResourceSummaryHandler: convert of string to list of strings

fstagni:


fstagni:
Ping! Anyone?

src/WebAppDIRAC/WebApp/handler/ResourceSummaryHandler.py
```diff
@@ -312,6 +310,8 @@ def web_action(self, action=None, **kwargs):
 
     def web_getResourceSummaryData(self, name=None, status=None, elementType=None, statusType=None, tokenOwner=None):
         """This method returns the data required to fill the grid."""
+        if isinstance(elementType, str):  # it is passed in as e.g. '["Catalog"]' 
+            elementType = eval(elementType)
```
andresailer:
```suggestion
            elementType = [ele.strip("\"' ") for ele in elementType.strip("[]").split(",")]
```

chrisburr: Is this being passed as `json`? If so it should be `json.loads(elementType)` else failing that `ast.literal_eval` would be a better choice.

fstagni: Updated to use `json` (seems like it's done elsewhere too)

src/WebAppDIRAC/WebApp/handler/ResourceSummaryHandler.py
```diff
@@ -312,6 +311,17 @@ def web_action(self, action=None, **kwargs):
 
     def web_getResourceSummaryData(self, name=None, status=None, elementType=None, statusType=None, tokenOwner=None):
         """This method returns the data required to fill the grid."""
+        if name:
+            name = list(json.loads(name))
```
chrisburr: What is the `list()` doing here? I struggle to see how could be different to just:
```suggestion
            name = json.loads(name)
```
(same for the later lines)

PR #699, opened 2022-10-17
[5.0] Running webapp without using nginx

chrisburr:
Leaving as draft as this needs to be thought about carefully.

BEGINRELEASENOTES

FIX: Running webapp without using nginx

ENDRELEASENOTES

### DIFF
--- a/src/WebAppDIRAC/Core/App.py
+++ b/src/WebAppDIRAC/Core/App.py
@@ -133,7 +133,7 @@ class App:
                 keyfile=Conf.HTTPSKey(),
                 cert_reqs=ssl.CERT_OPTIONAL,
                 ca_certs=Conf.generateCAFile(),
-                ssl_version=ssl.PROTOCOL_TLSv1_2,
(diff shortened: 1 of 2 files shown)


chaen:
seems ok to me

chaen (approved):

fstagni:
Well, this is something that we said to "maybe not merge" at BiLD meetings. Opinion?

chaen:
why not ? And what would the solution be then ?

fstagni:
The solution would be "install nginx", at least for production setup. For testing it would be useful...

Anyway, @chrisburr was not sure about the correctness of this one.

chaen:
I see. 
Well, we can definitely recommend to have nginx for production setup, but I can't think of anything wrong with this patch

marianne013:
Would we be able to put that in the next release ? Judging from Luisa's comments CTA also runs without nginx. 
We are currently running off this  branch, but it's not a permanent solution, especially if Janusz gets creative :-D

PR #701, opened 2022-10-27
[5.1] DIRACScript replaced by DIRAC.Core.Base.Script and remove VMDiracHandler

fstagni:
BEGINRELEASENOTES

FIX: DIRACScript replaced by DIRAC.Core.Base.Script
FEAT: Remove VMDiracHandler

ENDRELEASENOTES

Is "tornado-start-web" needed or it could be removed?


chrisburr:
> Is "tornado-start-web" needed or it could be removed?

I have no idea, I wonder if we're supposed to be using it instead of `dirac-webapp-run` when running without nginx.

fstagni:
> > Is "tornado-start-web" needed or it could be removed?
> 
> I have no idea, I wonder if we're supposed to be using it instead of `dirac-webapp-run` when running without nginx.

No idea either. Maybe @sfayer knows...?

sfayer:
We don't appear to use it anywhere in gridpp, so I suspect it's unused...

chrisburr:
`tornado-start-web` was new in v5 and was added as part of the oauth stuff (https://github.com/DIRACGrid/WebAppDIRAC/pull/447) so I wouldn't expect it to be used anywhere yet.

chrisburr (approved):

PR #705, opened 2022-11-10
[v5r0] fix Accounting option "Do not scale units"

aldbr:
This PR should resolve the issue we got during the hackathon with the "Do not scale units".

BEGINRELEASENOTES
FIX: DoNotScaleUnits option from the Accounting dashboard
ENDRELEASENOTES

### DIFF
--- a/src/WebAppDIRAC/WebApp/handler/AccountingHandler.py
+++ b/src/WebAppDIRAC/WebApp/handler/AccountingHandler.py
@@ -153,9 +153,9 @@ class AccountingHandler(WebHandler):
         del pD["timeSelector"]
 
         # Find extra parameters
-        for k in pD:
+        for k in list(pD.keys()):
             if k.find("ex_") == 0:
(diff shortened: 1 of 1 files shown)


src/WebAppDIRAC/WebApp/handler/AccountingHandler.py
```diff
@@ -153,9 +153,9 @@ def __parseFormParams(self, timeSelector, typeName=None, **pD):
         del pD["timeSelector"]
 
         # Find extra parameters
-        for k in pD:
+        for k in list(pD.keys()):
             if k.find("ex_") == 0:
-                extraParams[k[3:]] = pD[k]
+                extraParams[k[3:]] = pD.pop(k)
```
fstagni, line 158: why `pop`?

aldbr: To remove the key/value from `pD`.

PR #715, opened 2023-01-23
[v5.0] fix (WebSocketHandler): the web socket handler was not setting up the TC

chaen:
All CS and Registry were executed using the host certificate. VERY nice loophole :-) 

BEGINRELEASENOTES

FIX: WebSocketHandler sets the ThreadConfig before calling on_message

ENDRELEASENOTES

### DIFF
--- a/src/WebAppDIRAC/Lib/WebHandler.py
+++ b/src/WebAppDIRAC/Lib/WebHandler.py
@@ -369,3 +369,15 @@ class WebSocketHandler(tornado.websocket.WebSocketHandler, WebHandler):
     def _getMethod(self):
         """Get method function to call."""
         return self.on_open
+
+    def _on_message(self, msg):
+        """This needs to be implemented instead of ``on_message``"""
+        raise NotImplementedError('"_on_message" method is not implemented')
+
+    def on_message(self, msg):
+        """Setup the threadConfig before doing the actual calls.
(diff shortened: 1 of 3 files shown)


src/WebAppDIRAC/Lib/WebHandler.py
```diff
@@ -369,3 +369,11 @@ def on_open(self):
     def _getMethod(self):
         """Get method function to call."""
         return self.on_open
+
+    def on_message(self, msg):
+        """Setup the threadConfig before doing the actual calls.
+            Developper should implement ``_on_message`` instead
```
fstagni:
```suggestion
            Developer should implement ``_on_message`` instead
```
and missing an empty line below.

src/WebAppDIRAC/Lib/WebHandler.py
```diff
@@ -369,3 +369,15 @@ def on_open(self):
     def _getMethod(self):
         """Get method function to call."""
         return self.on_open
+
+    def _on_message(self, msg):
+        """This needs to be implemented in stead of ``on_message``"""
```
chrisburr:
```suggestion
        """This needs to be implemented instead of ``on_message``"""
```

chrisburr (approved):

PR #748, opened 2023-12-01
feat: add OwnerGroup to be displayed as a default option

marianne013:
BEGINRELEASENOTES
NEW: Added OwnerGroup as a default to JobMonitor to account for the fact that DIRAC will soon be inherently multi-VO and hence the OwnerGroup of the Submitter will matter.
ENDRELEASENOTES


marianne013:
I want an override for that pylint failure. I made a change to a js class for goodness sake !

fstagni:
Wait a sec: https://github.com/DIRACGrid/WebAppDIRAC/pull/749

PR #755, opened 2024-02-09
Pilots simplification

fstagni:
Reflecting changes in DIRAC DB

### DIFF
--- a/src/WebAppDIRAC/WebApp/static/DIRAC/PilotMonitor/classes/PilotMonitor.js
+++ b/src/WebAppDIRAC/WebApp/static/DIRAC/PilotMonitor/classes/PilotMonitor.js
@@ -94,9 +94,6 @@ Ext.define("DIRAC.PilotMonitor.classes.PilotMonitor", {
       name: "GridType",
     },
     {
-      name: "TaskQueueID",
-    },
-    {
       name: "CurrentJobID",
     },
     {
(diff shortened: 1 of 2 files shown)


src/WebAppDIRAC/WebApp/static/DIRAC/PilotMonitor/classes/PilotMonitor.js
```diff
@@ -187,18 +181,6 @@ Ext.define("DIRAC.PilotMonitor.classes.PilotMonitor", {
       computingElement: "Computing Element",
       VO: "VO",
       owner: "Owner",
-      broker: "Broker",
-    };
-
-    var textFields = {
```
aldbr, line 193: Seems like `textFields` is still used in the module:

https://github.com/DIRACGrid/WebAppDIRAC/blob/6d6ca9aac0c167b8479f7ca25ef08d4b525bee51/src/WebAppDIRAC/WebApp/static/DIRAC/PilotMonitor/classes/PilotMonitor.js#L197

Also, did you delete `pilotId` on purpose?

aldbr (approved):

src/WebAppDIRAC/WebApp/static/DIRAC/PilotMonitor/classes/PilotMonitor.js
fstagni: Good catch! fixed.

PR #771, opened 2025-01-15
Allow pre-releases and fix error message in DIRAC version check

maxnoe:
BEGINRELEASENOTES
FIX: Allow pre-releases in the DIRAC version check when starting the WebAPP and allow pre-releases for development
ENDRELEASENOTES

### DIFF
--- a/.github/workflows/deployment.yml
+++ b/.github/workflows/deployment.yml
@@ -29,7 +29,7 @@ jobs:
       - name: Checkout main repository
         uses: actions/checkout@v5
         with:
-          fetch-depth: 0  # get full history (so no need for "git fetch --unshallow")
+          fetch-depth: 0 # get full history (so no need for "git fetch --unshallow")
 
       # Configure git identity
       - name: Configure git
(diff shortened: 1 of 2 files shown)


aldbr:
In the next few days, I will try to run the branch in certification, if it does not break, then I will merge it, thanks!

fstagni:
Given that we will not have anymore pre-releases (ever!), I do not think this is needed. Re-open if you think I am wrong.

maxnoe:
> Given that we will not have anymore pre-releases (ever!), I do not think this is needed. Re-open if you think I am wrong.

The motivation for this change was local development! Where the version is based on the last tag plus some `.devXXX` git information, aka a pre-release.

It also fixes a genuine bug in the error message.

maxnoe:
At least the fix for the error message should be taken. I can split if you want, but I think this should go in

fstagni:
> > Given that we will not have anymore pre-releases (ever!), I do not think this is needed. Re-open if you think I am wrong.
> 
> The motivation for this change was local development! Where the version is based on the last tag plus some `.devXXX` git information, aka a pre-release.
> 
> It also fixes a genuine bug in the error message.

Ah, sorry about that. Re-opened then.

aldbr (approved):

PR #772, opened 2025-02-13
Fixes v9

fstagni:


src/WebAppDIRAC/WebApp/handler/ProxyManagerHandler.py
```diff
@@ -72,9 +71,9 @@ def web_getProxyManagerData(
             proxies.append(
                 {
                     "proxyid": f"{record[1]}@{record[2]}",
```
aldbr: IIUC, `record[2]` was the `UserGroup` , which is not available anymore.
Therefore, we probably want to remove it from the `proxyid`, don't we?

```suggestion
                    "proxyid": f"{record[1]}",
```

aldbr (approved):

PR #791, opened 2026-04-14
fix: Only auto-reload webapp in development mode

sfayer:
Hi,

The webapp has autoreload enabled if there is only one process: This causes it to stat all of the loaded files (hundreds of them, including all of the python site-package libraries) every second to see if they've changed. In the interests of efficiency, I'd like to suggest that this feature is only enabled in development mode (does anyone actually rely on it for their production instances?). Adding @marianne013 as a watcher.

Regards
Simon

BEGINRELEASENOTES
CHANGE: Only auto-reload webapp code in debug mode
ENDRELEASENOTES

### DIFF
--- a/src/WebAppDIRAC/Core/App.py
+++ b/src/WebAppDIRAC/Core/App.py
@@ -96,12 +96,14 @@ class App:
         SessionData.setHandlers(self.__handlerMgr.getHandlers()["Value"])
         # Create the app
         tLoader = TemplateLoader(self.__handlerMgr.getPaths("template"))
+        debug = Conf.devMode()
+        autoreload = (Conf.numProcesses() < 2) and debug
         kw = dict(
-            debug=Conf.devMode(),
+            debug=debug,
             template_loader=tLoader,
             cookie_secret=str(Conf.cookieSecret()),
             log_function=self._logRequest,
-            autoreload=Conf.numProcesses() < 2,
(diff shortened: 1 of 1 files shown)


src/WebAppDIRAC/Core/App.py
```diff
@@ -96,12 +96,14 @@ def bootstrap(self):
         SessionData.setHandlers(self.__handlerMgr.getHandlers()["Value"])
         # Create the app
         tLoader = TemplateLoader(self.__handlerMgr.getPaths("template"))
+        debug = conf.devMode()
```
fstagni:
```suggestion
        debug = Conf.devMode()
```

src/WebAppDIRAC/Core/App.py
```diff
@@ -96,12 +96,14 @@ def bootstrap(self):
...
+        autoreload = (conf.numProcesses() < 2) and debug
```
fstagni:
```suggestion
        autoreload = (Conf.numProcesses() < 2) and debug
```

fstagni (changes requested):

sfayer:
Ah yes, that's what I get for manually typing out changes from the test node rather than copying a proper patch file :-)
Should all be fixed now...

Regards,
Simon

fstagni (approved):

